PR #11, opened 2017-06-27
Tweak display of FeatureCollection layers

mhdostal:
This will fix the issue with duplicate layer names appearing for Feature Collection layers with only one sublayer.  For FC layers with more than one sublayer there will be no change


mhdostal:
Are we going to review and merge this before tomorrow's release?

cc: @dg0yal

PR #21, opened 2017-12-21
Compass

mhdostal:
Updated Compass element (formerly NorthArrow) to match new design.


Toolkit/ArcGISToolkit/Compass.swift
```diff
@@ -0,0 +1,105 @@
...
+        }
+    }
+    public var autoHide: Bool = true { // Auto hides when north is up
+        didSet {
+            animateCompass()
+        }
+    }
+    public var width: CGFloat = 30.0 {
+        didSet {
+            widthAnchor.constraint(equalToConstant: width).isActive = true
```
rolson: do you have to disable the previous widthAnchor constraint?

PR #22, opened 2017-12-22
Compass

mhdostal:
Added Compass component which can be added to the map view to see the current rotation and allow resetting the map rotation to 0 (by tapping the compass).


rolson:
@dg0yal please review for update 2

mhdostal:
@dg0yal - do you want to review this before it goes in?

PR #25, opened 2018-02-28
New Component: Time Slider

njarecha:
@mhdostal @rolson Please review.


rolson:
@mhdostal do you think we should create a "develop" branch and have this PR go into "develop" instead of "vNext"? I think we need to move away from the vNext branch because of the organizational structure that it was from the beginning, and we can't just merge vNext into master at some point. We have to merge components one by one.

mhdostal:
@rolson What you're describing for the "develop" branch sounds a lot like what I thought "vNext" was, but maybe I'm not clear on what "vNext" was supposed to be.

If we want to merge components one by one, then wouldn't each component need to be in it's own branch?  Merging components one-by-one might be hard too, because there will be merge conflicts in the .xcodeproj file (from past experience).

If we have a plan for what we want in the next release, then yes, I see the advantages of a "develop" branch.  Everything approved for the next release would go into that, then when we "release" it, we merge it into master, update the tags, and make sure the doc is up-to-date.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2507 @@
...
+                    }
+                })
+            }
+            
+            dispatchGroup.notify(queue: DispatchQueue.main, execute: {
+                //
+                // If full extent or time step interval is not available then
+                // we cannot initialize the slider. Finish with error.
+                if timeAwareLayersFullExtent == nil || timeAwareLayersStepInterval == nil {
+                    completion(NSError(domain: "", code: AGSErrorCode.commonFileNotFound.rawValue, userInfo: [NSLocalizedDescriptionKey : "There are no time aware layers to initialize time slider."]))
```
mhdostal: Instead of `commonFileNotFound`, use `commonNoData`

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2507 @@
...
+                // return with an error.
+                guard error == nil else {
+                    completion(error)
+                    return
+                }
+                
+                // The layer must support time filtering and it should enabled
+                // then only we can initialize the time slider.
+                if timeAwareLayer.supportsTimeFiltering && timeAwareLayer.isTimeFilteringEnabled {
+                    //
```
mhdostal: remove extra lines

Examples/ArcGISToolkitExamples/TimeSliderExample.swift
```diff
@@ -0,0 +1,89 @@
...
+        mapImageLayer.load(completion: { [weak self] (error) in
+            guard error == nil else {
+                print("error: \(String(describing: error))")
+                return
+            }
+            
+            // Zoom to full extent of layer
+            self?.mapView.setViewpoint(AGSViewpoint(targetExtent: mapImageLayer.fullExtent!), completion: nil)
+            
+            self?.timeSlider?.initializeTimeProperties(timeAwareLayer: mapImageLayer, completion: { [weak self] (error) in
```
mhdostal: what's the difference between this method and the one that takes an `AGSGeoView`?  Why can't you just use the `geoView` method here?  (meaning, do we really need both?)

rolson:
If you use an error code from the AGSErrorCode, you need to use the matching domain. Error codes and domains are tied to the hip

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2507 @@
...
+                // With the given positioning of major tick marks, check whether their labels will overlap.
+                for j in stride(from: firstMajorTickIndex, to: tickCount - prospectiveInterval, by: i) {
+                    //
+                    // Get the current tick and it's label
+                    let currentTick = tms[j]
+                    var currentTickLabelFrame: CGRect? = nil
+                    if let currentTickDate = currentTick.value, let currentTickXPosition = currentTick.originX {
+                        let currentTickLabelString = timeStepIntervalDateFormatter.string(from: currentTickDate)
+                        let currentTickLabel = tickMarkLabel(with: currentTickLabelString, originX: currentTickXPosition)
+
```
mhdostal: this would be more readable if the function name was simply `changeContainsTimeAwareLayer`.

mhdostal: Do we need this in the production code?

mhdostal: Can the `fullExtentEndTime` and `fullExtentEndTime` ever be the same?  If so, this line will cause problems with divide by zero.

mhdostal: This check is present in the if statements below, so I don't think it's really necessary.

mhdostal: Is it possible to break out of a `forEach` loop?  Once you find a time aware layer, you don't need to continue...

mhdostal: ... or maybe just `return true` instead of setting `isTimeAware = true`

mhdostal: You use this value (6) in the next method, so these should be constants.

mhdostal: Constant...

mhdostal: use constant...

Examples/ArcGISToolkitExamples/TimeSliderExample.swift
```diff
@@ -0,0 +1,89 @@
...
+        timeSlider.translatesAutoresizingMaskIntoConstraints = false
+        timeSlider.bottomAnchor.constraint(equalTo: mapView.attributionTopAnchor, constant: -margin).isActive = true
+        timeSlider.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: margin).isActive = true
+        timeSlider.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: -margin).isActive = true
+        
+        // Set time slider 
+        self.timeSlider = timeSlider
+        
+        // Add layer
+        let mapImageLayer = AGSArcGISMapImageLayer(url: URL(string: "https://rtc-100-3.esri.com/arcgis/rest/services/timeSupport/cycle_routes/MapServer")!)
```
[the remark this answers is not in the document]
njarecha: Which public layer we are going to use is not decided yet so using this layer to show the slider working. It will change in the end.

Examples/ArcGISToolkitExamples/TimeSliderExample.swift
```diff
@@ -0,0 +1,89 @@
...
+        mapView.map?.operationalLayers.add(mapImageLayer)
+        mapImageLayer.load(completion: { [weak self] (error) in
+            guard error == nil else {
+                print("error: \(String(describing: error))")
+                return
+            }
+            
+            // Zoom to full extent of layer
+            self?.mapView.setViewpoint(AGSViewpoint(targetExtent: mapImageLayer.fullExtent!), completion: nil)
```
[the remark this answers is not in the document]
njarecha: fixed

Examples/ArcGISToolkitExamples/TimeSliderExample.swift
```diff
@@ -0,0 +1,89 @@
...
+                defer {
+                    self?.timeSlider?.isHidden = false
+                }
+            })
+        })
+    }
+    
+    @objc func timeSliderValueChanged(timeSlider: TimeSlider) {
+        print("Time slider value changed: (\(String(describing: timeSlider.currentExtent?.startTime)), \(String(describing: timeSlider.currentExtent?.endTime)))")
+        if mapView.timeExtent != timeSlider.currentExtent {
```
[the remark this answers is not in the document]
njarecha: I had the same question regarding `currentExtent` vs `currentTimeExtent` but I went as per design. Since, it is a TimeSlider, it does not make sense to use `Time` in all properties. So I did not raised that question.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2507 @@
...
+public class TimeSlider: UIControl {
+    
+    // MARK: - Public Properties
+    
+    // MARK: Current Extent Properties
+    
+    /*
+     The current time window or time instant of the slider
+     */
+    public var currentExtent: AGSTimeExtent? {
```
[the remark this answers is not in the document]
njarecha: same comment as above.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2507 @@
...
+    case forward
+    case backward
+}
+
+/*
+ The looping behavior of the slider when time is being animated.
+ */
+public enum LoopMode {
+    case none
+    case `repeat`
```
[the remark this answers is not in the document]
njarecha: Because it is a key word.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2507 @@
...
+                //
+                // Update current extent start and end time
+                updateCurrentExtentStartTime(startTime)
+                updateCurrentExtentEndTime(endTime)
+                
+                // Update flag only if it's not an internal update
+                if !internalUpdate {
+                    //
+                    // Range is enabled only if both times are not same.
+                    isRangeEnabled = startTime == endTime ? false : true
```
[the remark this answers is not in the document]

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2507 @@
...
+                }
+            }
+            // This means there is only one thumb needs to be displayed and current extent start and end times are same.
+            else if let startTime = currentExtent?.startTime, currentExtent?.endTime == nil {
```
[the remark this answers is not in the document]
njarecha: I think this is more readable and easy to understand. Also, the `else if let` with `logical OR` is making the statement more complex.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2507 @@
...
+                
+                // Update current extent start time
+                updateCurrentExtentStartTime(endTime)
+
+                // Start and end time must be same.
+                currentExtentEndTime = currentExtentStartTime
+            }
+            
+            // Refresh the view
+            refresh()
```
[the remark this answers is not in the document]
njarecha: Yes, it should refresh so the thumbs will be on the ends of the slider.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2507 @@
...
+            }
+        }
+    }
+    
+    // MARK: Full Extent Properties
+    
+    /*
+     The full extent time window on the slider.
+     */
+    public var fullExtent: AGSTimeExtent? {
```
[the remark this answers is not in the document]
njarecha: See my explanation above.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2507 @@
...
+            //
+            // Set current extent if it's nil.
+            if currentExtent == nil, let fullExtent = fullExtent {
+                currentExtent = fullExtent
+            }
+            else {
+                //
+                // It is possible that the current extent times are outside of the range of
+                // new full extent times. Adjust and sanp them to the tick marks.
+                if let startTime = currentExtentStartTime, let endTime = currentExtentEndTime {
```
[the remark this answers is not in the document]

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2507 @@
...
+    /*
+     The size of the thumb.
+     */
+    public var thumbSize: CGSize = CGSize(width: 25.0, height: 25.0) {
+        didSet {
+            //
+            // More than 50 width and height
+            // is too big for the iOS control.
+            // So, restrict it.
+            if thumbSize.width > 50.0 {
```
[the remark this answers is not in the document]

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2507 @@
...
+            upperThumbLayer.setNeedsDisplay()
+        }
+    }
+    
+    // MARK: Playback Properties
+    
+    /*
+     A Boolean value indicating whether playback buttons are visible or not. Default is true.
+     */
+    public var isPlaybackButtonsVisible: Bool = true {
```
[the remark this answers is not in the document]
njarecha: Shall I just change it to `playbackButtonsVisible`?

Toolkit/ArcGISToolkit/TimeSlider.swift

Examples/ArcGISToolkitExamples/TimeSliderExample.swift
njarecha: There are can be more than one time aware layers in a map. If you want time slider just for specific layer then you are going to use this method. If you want time slider for all layers in a map then you are going to use which takes `geoView`.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2507 @@
...
+        fullExtentEndTimeLabel.isHidden = !fullExtentLabelsVisible
+        fullExtentEndTimeLabel.foregroundColor = fullExtentLabelColor.cgColor
+        fullExtentEndTimeLabel.alignmentMode = kCAAlignmentCenter
+        fullExtentEndTimeLabel.frame = CGRect.zero
+        fullExtentEndTimeLabel.contentsScale = UIScreen.main.scale
+        fullExtentEndTimeLabel.fontSize = fullExtentLabelFont.pointSize
+        layer.addSublayer(fullExtentEndTimeLabel)
+        
+        // Add the lower value label
+        currentExtentStartTimeLabel.isHidden = labelMode == .thumbs ? true : false && isSliderVisible ? true : false
```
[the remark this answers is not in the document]
njarecha: It should be `!(labelMode == .thumbs && isSliderVisible)`

Toolkit/ArcGISToolkit/TimeSlider.swift

njarecha: removed

njarecha: Shall we show error in alert? or printing is fine?

njarecha:
@mhdostal I've addressed all your comments. Please re-review.

Toolkit/ArcGISToolkit/TimeSlider.swift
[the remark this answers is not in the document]
mhdostal: Yes, `playbackButtonsVisible` works for me.

Toolkit/ArcGISToolkit/TimeSlider.swift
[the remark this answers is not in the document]
mhdostal: Agreed.  I like `!(labelMode == .thumbs && isSliderVisible)` better than the original code.

PR #33, opened 2018-05-03
New Component: Time Slider

njarecha:
@rolson @esreli  Please review.


Examples/ArcGISToolkitExamples/TimeSliderExample.swift
```diff
@@ -0,0 +1,91 @@
...
+// limitations under the License.
+
+import UIKit
+import ArcGISToolkit
+import ArcGIS
+
+class TimeSliderExample: MapViewController {
+    
+    private var map : AGSMap?
+    private var timeSlider : TimeSlider?
```
esreli: I would suggest to explicitly init the TimeSlider from line 32 here, there seems to be no need for the timeSlider property to be optional.

Examples/ArcGISToolkitExamples/TimeSliderExample.swift
```diff
@@ -0,0 +1,91 @@
...
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        // Create an instance of a map with ESRI topographic basemap
+        self.map = AGSMap(basemap: AGSBasemap.topographic())
```
esreli: as per swift style guide, no need for `self`

Examples/ArcGISToolkitExamples/TimeSliderExample.swift
```diff
@@ -0,0 +1,91 @@
...
+        self.mapView.map = self.map
```
esreli: as per swift style guide, no need for `self`

Examples/ArcGISToolkitExamples/TimeSliderExample.swift
```diff
@@ -0,0 +1,91 @@
...
+            
+        // Add constraints to position the slider
+        let margin: CGFloat = 10.0
+        timeSlider.translatesAutoresizingMaskIntoConstraints = false
+        timeSlider.bottomAnchor.constraint(equalTo: mapView.attributionTopAnchor, constant: -margin).isActive = true
+        timeSlider.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: margin).isActive = true
+        timeSlider.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: -margin).isActive = true
+        
+        // Set time slider 
+        self.timeSlider = timeSlider
```
esreli: as per swift style guide, no need for `self`

Examples/ArcGISToolkitExamples/TimeSliderExample.swift
```diff
@@ -0,0 +1,91 @@
...
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        // Create an instance of a map with ESRI topographic basemap
+        self.map = AGSMap(basemap: AGSBasemap.topographic())
+        self.mapView.map = self.map
+        
+        // Configure time slider
+        let timeSlider = TimeSlider()
```
esreli: no need for this line if change on line 22 is honored

Examples/ArcGISToolkitExamples/TimeSliderExample.swift
```diff
@@ -0,0 +1,91 @@
...
+        timeSlider.isHidden = true
+        timeSlider.addTarget(self, action: #selector(TimeSliderExample.timeSliderValueChanged(timeSlider:)), for: .valueChanged)
+        view.addSubview(timeSlider)
+            
+        // Add constraints to position the slider
+        let margin: CGFloat = 10.0
+        timeSlider.translatesAutoresizingMaskIntoConstraints = false
+        timeSlider.bottomAnchor.constraint(equalTo: mapView.attributionTopAnchor, constant: -margin).isActive = true
+        timeSlider.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: margin).isActive = true
+        timeSlider.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: -margin).isActive = true
```
esreli: this anchor will need to be aware of the safe area's trailing anchor

Examples/ArcGISToolkitExamples/TimeSliderExample.swift
```diff
@@ -0,0 +1,91 @@
...
+        let timeSlider = TimeSlider()
+        timeSlider.isHidden = true
+        timeSlider.addTarget(self, action: #selector(TimeSliderExample.timeSliderValueChanged(timeSlider:)), for: .valueChanged)
+        view.addSubview(timeSlider)
+            
+        // Add constraints to position the slider
+        let margin: CGFloat = 10.0
+        timeSlider.translatesAutoresizingMaskIntoConstraints = false
+        timeSlider.bottomAnchor.constraint(equalTo: mapView.attributionTopAnchor, constant: -margin).isActive = true
+        timeSlider.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: margin).isActive = true
```
esreli: this anchor will need to be aware of the safe area's leading anchor

Examples/ArcGISToolkitExamples/TimeSliderExample.swift
esreli: if we remove line 32, this line is no longer needed

Examples/ArcGISToolkitExamples/TimeSliderExample.swift
```diff
@@ -0,0 +1,91 @@
...
+        
+        // Set time slider 
+        self.timeSlider = timeSlider
+        
+        // Add layer
+        let mapImageLayer = AGSArcGISMapImageLayer(url: URL(string: "https://rtc-100-3.esri.com/arcgis/rest/services/timeSupport/cycle_routes/MapServer")!)
+        mapView.map?.operationalLayers.add(mapImageLayer)
+        mapImageLayer.load(completion: { [weak self] (error) in
+            guard error == nil else {
+                print("error: \(String(describing: error))")
```
esreli: error has not been unwrapped & I would suggest using the localized description property instead

esreli: `print("error: \(error!.localizedDescription)")`

Examples/ArcGISToolkitExamples/TimeSliderExample.swift
```diff
@@ -0,0 +1,91 @@
...
+                return
+            }
+            
+            // Zoom to full extent of layer
+            if let fullExtent = mapImageLayer.fullExtent {
+                self?.mapView.setViewpoint(AGSViewpoint(targetExtent:fullExtent.toBuilder().expand(byFactor: 2.5).toGeometry()), completion: nil)
```
esreli: Because this is an example, could we break out the envelope builder code into 3 separate lines?

Examples/ArcGISToolkitExamples/TimeSliderExample.swift
```diff
@@ -0,0 +1,91 @@
...
+            }
+            
+            self?.timeSlider?.initializeTimeProperties(timeAwareLayer: mapImageLayer, completion: { [weak self] (error) in
+                guard error == nil else {
+                    print("error: \(String(describing: error))")
```
esreli: same comment as line 52

Examples/ArcGISToolkitExamples/TimeSliderExample.swift
```diff
@@ -0,0 +1,91 @@
...
+                    self?.showError(error!)
+                    return
+                }
+                print("Time slider is successfully initialized!")
+                defer {
```
esreli: defer statement might not be needed since it lives at the end of the current scope anyway. Perhaps place this above the guard statement if you always want the defer statement to be executed at the end of the current scope. Otherwise, the defer is not needed.

Examples/ArcGISToolkitExamples/TimeSliderExample.swift
```diff
@@ -0,0 +1,91 @@
...
+                    self?.timeSlider?.isHidden = false
+                }
+            })
+        })
+    }
+    
+    @objc func timeSliderValueChanged(timeSlider: TimeSlider) {
+        print("Time slider value changed: (\(String(describing: timeSlider.currentExtent?.startTime)), \(String(describing: timeSlider.currentExtent?.endTime)))")
```
esreli: Using string describing on an optional variable will describe the variable as optional instead of as it's variable.

njarecha:
@esreli I've addressed your comments. Please re-review `TimeSliderExample` code.

Examples/ArcGISToolkitExamples/TimeSliderExample.swift
```diff
@@ -0,0 +1,96 @@
...
+        
+        //
+        // Add constraints to position the slider
+        let margin: CGFloat = 10.0
+        timeSlider.translatesAutoresizingMaskIntoConstraints = false
+        timeSlider.bottomAnchor.constraint(equalTo: mapView.attributionTopAnchor, constant: -margin).isActive = true
+        
+        if #available(iOS 11.0, *) {
+            let layoutMarginsGuide = view.layoutMarginsGuide
+            timeSlider.leadingAnchor.constraint(equalTo: layoutMarginsGuide.leadingAnchor).isActive = true
```
esreli: If we want to maintain the margin constant we had before, I would suggest using the view's `safeAreaLayoutGuide` over the view's `layoutMarginsGuide`. The layout margins guide is Apple's recommend content layout margin and as far as i'm aware we cannot change the constant value whereas the safe area layout guide is the layout guide that determines where in the screen is safe to interact with and thus will allow us to reapply our desired 10.0 pt margin.
```
timeSlider.leadingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.leadingAnchor, constant: margin).isActive = true
timeSlider.trailingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.trailingAnchor, constant: -margin).isActive = true
```

esreli:
@njarecha I will review TimeSlider when I have a substantial chunk of time to do so, i'd like to give it my full attention for an extended duration of time.

Examples/ArcGISToolkitExamples/TimeSliderExample.swift
njarecha: fixed

esreli:
I'm not really sure why the snippets github decided to post in the 'Conversation' thread are all 4 lines starting at 1543 but you should be able to view my reviews properly under 'Files Changes'.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+                // Check that the prospective interval between major ticks results in
+                // an equal number of minor ticks on both ends.
+                for m in stride(from: prospectiveInterval, to: tickCount, by: prospectiveInterval) {
+                    let totalNumberOfTicksOnEnds = tickCount - m + 1
+                    
+                    // If the total number of minor ticks on both ends (i.e. before and after the
+                    // first and last major ticks) is less than the major tick interval being tested, then we've
+                    // found the number of minor ticks that would be on the ends if we use this major tick interval.
+                    // If that total is divisible by two, then the major tick interval under test allows for an
+                    // equal number of minor ticks on the ends.
```
esreli: Is this comment unfinished?

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+ */
+public enum PlaybackDirection {
+    case forward
+    case backward
+}
+
+/*
+ The looping behavior of the slider when time is being animated.
+ */
+public enum LoopMode {
```
esreli: I would consider parameterizing LoopMode enum cases `repeat` and `reverse` with a PlaybackDirection. This way, you won't need to maintain a reference to both a LoopMode and PlaybackDirection variable. Example:

```swift
public enum LoopMode {
    case none
    case `repeat`(PlaybackDirection)
    case reverse(PlaybackDirection)
    
    var direction: PlaybackDirection? {
        switch self {
        case .none:
            return nil
        case .`repeat`(let dir), .reverse(let dir):
            return dir
        }
    }
}

public enum PlaybackDirection {
    case forward
    case backward
}

let mode = LoopMode.reverse(.forward)
```

esreli: This would also rid the need for a nested `switch` statement in line 1116.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+    
+    private static var KVOContext = 0
+    private var map: AGSMap?
+    private var scene: AGSScene?
+    private var isObserving: Bool = false
+    
+    // MARK: - Override Functions
+    
+    required public init?(coder: NSCoder) {
+        super.init(coder: coder)
```
esreli: What IB support do you plan on offering? I don't think this init implementation would properly setup this component from a storyboard or nib. I would either support IB or perform the default assert.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+        currentExtentEndTimeLabel.isHidden = !(labelMode == .thumbs && isSliderVisible)
+        currentExtentEndTimeLabel.foregroundColor = currentExtentLabelColor.cgColor
+        currentExtentEndTimeLabel.alignmentMode = kCAAlignmentCenter
+        currentExtentEndTimeLabel.frame = CGRect.zero
+        currentExtentEndTimeLabel.contentsScale = UIScreen.main.scale
+        currentExtentEndTimeLabel.fontSize = currentExtentLabelFont.pointSize
+        layer.addSublayer(currentExtentEndTimeLabel)
+        
+        // Create the images
+        let bundle = Bundle(for: type(of: self))
```
esreli: Constructing the following 4 images from bundle is unnecessary, no? Wouldn't `UIImage(named: "ImageName")!` for each of these 4 images suffice?

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+        
+        // Add forward button
+        let f = UIButton(type: .custom)
+        f.setBackgroundImage(forwardImage, for: .normal)
+        f.addTarget(self, action: #selector(TimeSlider.forwardAction(_:)), for: .touchUpInside)
+        f.showsTouchWhenHighlighted = true
+        addSubview(f)
+        forwardButton = f
+        
+        // Add forward button
```
esreli: change comment to read `// Add backward button`

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+     */
+    public private(set) var geoView: AGSGeoView?
+    
+    /*
+     A Boolean value indicating whether slider is visible or not. Default is true.
+     */
+    public var isSliderVisible: Bool = true {
+        didSet {
+            layoutIfNeeded()
+            UIView.animate(withDuration: 0.25) {
```
esreli: This animation seems unnecessary because the frame is adjusted without animation and thus the components that are becoming animated hidden are doing so after the frame has adjusted and thus looks strange. Consider wrapping the frame adjustment as an animation within this block as well or removing this animation block altogether.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+                
+                // Update current extent start time
+                updateCurrentExtentStartTime(endTime)
+                
+                // Start and end time must be same.
+                currentExtentEndTime = currentExtentStartTime
+            }
+            // Set start and end time to nil if
+            // current extent is nil
+            else if currentExtent == nil {
```
philium: @njarecha Is it possible for `currentExtent` to be non-`nil`, but for `startTime` and `endTime` to be `nil`? If so, it looks like that case might not be handled.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import UIKit
+import QuartzCore
+import ArcGIS
+
+// MARK: - Time Slider Control
+
+/*
```
philium: @njarecha You could make these comments viewable in Xcode's quick help by adding a second `*`, or by using `///`. The easiest way to make these documentation comments is with the Add Documentation (⌥⌘/) keyboard shortcut. That shortcut is extremely helpful for documenting functions as it adds placeholders for parameters and the return value.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+ Controls how labels on the slider are displayed.
+ */
+public enum TimeSliderLabelMode {
```
philium: @njarecha Could these types (`TimeSliderLabelMode`, `PlaybackDirection`, `LoopMode`, `PlaybackDirection`, and `Theme`) be nested under `TimeSlider`? They are documented as being specific to a time slider, but they are currently in the global namespace.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+    public var fullExtentBorderColor: UIColor = UIColor.black {
+        didSet {
+            trackLayer.setNeedsDisplay()
+        }
+    }
+    
+    /*
+     The border width of the full extent of the slider.
+     */
+    public var fullExtentBorderWidth: Double = 1.0 {
```
philium: It might make more sense for this to be a `CGFloat`. This seems to be a drawing-related property and making it a CGFloat would prevent having to cast it later while drawing.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+            tickMarkLayer.setNeedsDisplay()
+        }
+    }
+    
+    /*
+     Controls how labels on the slider are displayed. The default is thumbs.
+     */
+    public var labelMode: TimeSliderLabelMode = .thumbs {
+        didSet {
+            switch labelMode {
```
philium: It looks like the content of the first and third cases are identical, so they could be combined using `case .none, .ticks:`.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+    private var isObserving: Bool = false
+    
+    // MARK: - Override Functions
+    
+    required public init?(coder: NSCoder) {
+        super.init(coder: coder)
+    }
+    
+    override public var frame: CGRect {
+        didSet {
```
philium: If I recall correctly, this property observer will not always be triggered by changes to the view's bounds, even though the frame would necessarily change. Are you wanting to redraw the view when the size changes? Consider setting the view's `contentMode` to `redraw`. If you need to perform an action when the view's size changes, a more reliable way is to query the `frame` in `layoutSubviews()`.

Toolkit/ArcGISToolkit/TimeSlider.swift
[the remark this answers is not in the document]
philium: Or make `AGSTimeValue` conform to `Comparable`.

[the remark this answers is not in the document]
philium: Another Swifty alternative is to add an initializer to `TimeInterval` that takes an `AGSTimeValue`:

```
extension TimeInterval {
    init(timeValue: AGSTimeValue) {
        // …
    }
}
```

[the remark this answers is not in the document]
esreli: @philium wouldn't your suggested extension initializer need to be qualified with the `convenience` keyword?

esreli:
I found a bug where if slider is set to visible false and playback buttons visible are set to false and then I change the Thumb Options to T Win, a thumb button is drawn to screen.

screen shot 2018-05-08 at 11 08 46 am

Toolkit/ArcGISToolkit/TimeSlider.swift
esreli: Is it at all possible for this to be NaN? (albeit extremely unlikely) Should you check for NaN? I'm always wary when using division with floating point types.

esreli: ditto comment from line 1704

esreli: is it at all possible to cause a fatal error? i.e. initializing an `Int` with a `Double` that's greater than `Int.max`?

esreli: I recognize this is an absurd edge case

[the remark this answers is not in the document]
philium: @esreli Good question. The `convenience` keyword is not needed because `TimeInterval` (`Double`) is a struct. That keyword only makes sense in the context of a class. From the [language guide](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Initialization.html#//apple_ref/doc/uid/TP40014097-CH18-ID216):

> Convenience initializers are secondary, supporting initializers for a class.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+        }
+        
+        // Commit the transaction
+        CATransaction.commit()
+    }
+    
+    private func positionTickMarks() {
+        //
+        // Bail out if time steps are not available
+        guard let timeSteps = timeSteps, timeSteps.count > 0, isSliderVisible else {
```
philium: `!timeSteps.isEmpty` is preferred to `timeSteps.count > 0` as it is guaranteed to be O(1).

> When you need to check whether your collection is empty, use the isEmpty property instead of checking that the count property is equal to zero. For collections that don’t conform to RandomAccessCollection, accessing the count property iterates through the elements of the collection.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+            return
+        }
+        
+        // Create tick marks based on
+        // time steps and it's calculated
+        // position on the slider
+        var tms = [TickMark]()
+        for i in 0..<timeSteps.count  {
```
philium: While this `for-in` loop works, it could be improved upon. It assumes that the `startIndex` of `timeSteps` is `0`. That is true for `Array`, but not true for all types that conform to `Collection`. Rather than constructing the index range manually, it is better to use the `indices` property of `Collection`: `for i in timeSteps.indices { … }`. And if you don't actually need the index, which is the case here, you could just use `for timeStep in timeSteps { … }`.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+        // add labels for major ticks
+        if labelMode == .ticks {
+            var majorTickInterval = 2
+            var doMajorTickLabelsCollide = false
+            var firstMajorTickIndex = 0
+            let tickCount = tms.count
+            
+            // Calculate the largest number of ticks to allow between major ticks. This prevents scenarios where
+            // there are two major ticks placed undesirably close to the end.
+            let maxMajorTickInterval = Int(ceil(Double(tickCount / 2)))
```
philium: I think you intended to cast `tickCount` to `Double` before dividing by `2`. The way it is now, both `tickCount` and `2` are integers, so integer division will be used. The value of `maxMajorTickInterval` would just be the quotient.

Toolkit/ArcGISToolkit/TimeSlider.swift
philium: Is it possible for there to be only one tick? If so, this would result in a runtime error as well.

One alternative would be to drop the first and last indexes:

```
for i in indices.dropFirst().dropLast() {
   …
}
```

philium: This is a rather unusual use of `guard`. It isn't really guarding against anything because the closure returns right after the `guard` statement. An `if` statement might be a bit more straightforward.

```
if let error = error {
    print("error: \(error)")
}
```

Toolkit/ArcGISToolkit/TimeSlider.swift

Toolkit/ArcGISToolkit/TimeSlider.swift

Toolkit/ArcGISToolkit/TimeSlider.swift
njarecha: done

Toolkit/ArcGISToolkit/TimeSlider.swift

Toolkit/ArcGISToolkit/TimeSlider.swift

Toolkit/ArcGISToolkit/TimeSlider.swift

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+        didSet {
+            lowerThumbLayer.setNeedsDisplay()
+            upperThumbLayer.setNeedsDisplay()
+        }
+    }
+    
+    /*
+     The size of the thumb.
+     */
+    public var thumbSize: CGSize = CGSize(width: 25.0, height: 25.0) {
```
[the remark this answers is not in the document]

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+    
+    // MARK: Playback Properties
+    
+    /*
+     A Boolean value indicating whether playback buttons are visible or not. Default is true.
+     */
+    public var playbackButtonsVisible: Bool = true {
+        didSet {
+            layoutIfNeeded()
+            UIView.animate(withDuration: 0.25) {
```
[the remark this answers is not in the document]
njarecha: `UIView.transition(with: self, duration: 0.5, options: .transitionCrossDissolve, animations: ` works so using it now.

Toolkit/ArcGISToolkit/TimeSlider.swift
njarecha: same comment as above.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+    /*
+     Different color themes of the slider.
+     */
+    public var theme: Theme = .black {
+        didSet {
+            //
+            // Set the color based on
+            // selected option
+            var tintColor: UIColor
+            var backgroundColor: UIColor
```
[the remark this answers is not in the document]

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+    // If set to True, it will show two thumbs, otherwise only one. Default is True.
+    fileprivate var isRangeEnabled: Bool = true {
+        didSet {
+            upperThumbLayer.isHidden = !isRangeEnabled
+            currentExtentEndTimeLabel.isHidden = !isRangeEnabled
+        }
+    }
+    
+    private var heightConstraint: NSLayoutConstraint?
+    private var timer = Timer()
```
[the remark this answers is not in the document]

Toolkit/ArcGISToolkit/TimeSlider.swift
njarecha: throwing a fatalError now.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+    private var scene: AGSScene?
+    private var isObserving: Bool = false
+    
+    // MARK: - Override Functions
+    
+    required public init?(coder: NSCoder) {
+        super.init(coder: coder)
+    }
+    
+    override public var frame: CGRect {
```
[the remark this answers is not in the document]

Toolkit/ArcGISToolkit/TimeSlider.swift
njarecha: Setting `contentMode` to `redraw` and `refresh` is already being called from `layoutSubviews()`.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+        setupUI()
+    }
+    
+    override public func beginTracking(_ touch: UITouch, with event: UIEvent?) -> Bool {
+        //
+        // Stop playing
+        isPlaying = false
+        
+        // Set preview location from touch
+        previousLocation = touch.location(in: self)
```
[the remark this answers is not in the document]
njarecha: change it to `location` and removed `previousLocation`.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+        return lowerThumbLayer.isHighlighted || upperThumbLayer.isHighlighted
+    }
+    
+    open override func continueTracking(_ touch: UITouch, with event: UIEvent?) -> Bool {
+        //
+        // Get the touch location
+        let location = touch.location(in: self)
+        
+        // Update previous location
+        previousLocation = location
```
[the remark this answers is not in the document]
njarecha: removed `previousLocation`.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+     */
+    public func initializeTimeProperties(geoView: AGSGeoView, completion: @escaping (Error?)->Void) {
+        //
+        // Set geoview on the slider
+        // it will initialize required properties.
+        self.geoView = geoView
+        
+        //
+        // Get all layers from geoView
+        var operationalLayers = [AGSLayer]()
```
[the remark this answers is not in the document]

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+        // to be loaded to gather requried information
+        // to load time properties
+        let dispatchGroup = DispatchGroup()
+        
+        // Load all operational layers to initialize slider properties
+        AGSLoadObjects(operationalLayers) { [weak self] (loaded) in
+            //
+            // Once, all layers are loaded,
+            // loop through all of them.
+            if loaded {
```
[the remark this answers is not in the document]

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+                        //
+                        // Get the layer's full time extent and combine with other layer's full extent.
+                        if let fullTimeExtent = timeAwareLayer.fullTimeExtent {
+                            timeAwareLayersFullExtent = timeAwareLayersFullExtent == nil ? timeAwareLayer.fullTimeExtent : timeAwareLayersFullExtent?.union(otherTimeExtent: fullTimeExtent)
+                        }
+                        
+                        // This is an async operation to find out time step interval and
+                        // whether range time filtering is supported by the layer.
+                        dispatchGroup.enter()
+                        self?.findTimeStepIntervalAndIsRangeTimeFilteringSupported(for: timeAwareLayer, completion: { (arg0) in
```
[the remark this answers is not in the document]

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+                            if let (timeInterval, supportsRangeFiltering) = arg0 {
+                                supportsRangeTimeFiltering = supportsRangeFiltering
+                                
+                                // We are looking for the greater time interval than we already have it.
+                                if let timeInterval = timeInterval {
+                                    if var timeAwareLayersStepInterval = timeAwareLayersStepInterval {
```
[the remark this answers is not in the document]

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+                                        timeAwareLayersStepInterval = timeInterval.isGreaterThan(otherTimeValue: timeAwareLayersStepInterval) ? timeInterval : timeAwareLayersStepInterval
```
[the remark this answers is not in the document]

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+                        })
+                    }
+                })
+            }
+            
+            dispatchGroup.notify(queue: DispatchQueue.main, execute: {
+                //
+                // If full extent or time step interval is not available then
+                // we cannot initialize the slider. Finish with error.
+                if timeAwareLayersFullExtent == nil || timeAwareLayersStepInterval == nil {
```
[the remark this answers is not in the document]

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+                }
+            })
+        }
+    }
+    
+    /*
+     This will initialize slider's fullExtent, currentExtent and timeStepInterval properties
+     if the layer is visible and participate in time based filtering.
+     */
+    public func initializeTimeProperties(timeAwareLayer: AGSTimeAware, completion: @escaping (Error?)->Void) {
```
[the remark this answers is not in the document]
njarecha: The `initializeTimeProperties(geoView: AGSGeoView` deals with multiple layers to get the time enabled properties but this function `initializeTimeProperties(timeAwareLayer` deals with only one layer. Let's keep them separate it's easy to understand.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+        //
+        // The layer must be loadable.
+        if let layer = timeAwareLayer as? AGSLoadable {
```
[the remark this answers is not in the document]

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+                // If layer fails to load then
+                // return with an error.
+                guard error == nil else {
+                    completion(error)
+                    return
+                }
+                
+                // The layer must support time filtering and it should enabled
+                // then only we can initialize the time slider.
+                if timeAwareLayer.supportsTimeFiltering && timeAwareLayer.isTimeFilteringEnabled {
```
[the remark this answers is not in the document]

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+        else {
+            completion(NSError(domain: AGSErrorDomain, code: AGSErrorCode.commonNoData.rawValue, userInfo: [NSLocalizedDescriptionKey : "The layer is not loadable to initialize time slider."]))
+        }
+    }
+    
+    /*
+     This will re-calculate's timeStepInterval based on the provided step count.
+     The time slider must have full extent available to calcluate the timeStepInterval.
+     */
+    public func initializeTimeSteps(timeStepCount: Int) {
```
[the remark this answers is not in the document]
njarecha: There is already a design issue for this but for now just added `completion` so user will get an error.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+    }
+    
+    /*
+     Moves the slider thumbs back with provided time steps.
+     */
+    @discardableResult public func stepBack(timeSteps: Int) -> Bool {
+        //
+        // Time steps must be greater than 0
+        if timeSteps > 0 {
+            return moveTimeStep(timeSteps: 0 - timeSteps)
```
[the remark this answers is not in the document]

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+    }
+    
+    @objc private func timerAction() {
+        if let geoView = self.geoView {
+            if geoView.drawStatus == .completed {
+                handlePlaying()
+            }
+        }
+        else {
+            handlePlaying()
```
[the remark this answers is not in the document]
njarecha: Whether `geoView` is there or not we have to handle the `playing`. When geoView is there, time slider waits for the drawing to be completed before moving to the next time stamp.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+        backButton = b
+        
+        // Refresh
+        refresh()
+    }
+    
+    private func refresh() {
+        //
+        // Calculate time steps
+        if let timeSteps = self.timeSteps, timeSteps.isEmpty || self.timeSteps == nil {
```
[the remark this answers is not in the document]
njarecha: Changed condition. The `calculateTimeSteps()` is here for odd cases like if someone set's `fullExtent` and/or `currentExtent` without calling helper initialize methods.

Toolkit/ArcGISToolkit/TimeSlider.swift

Toolkit/ArcGISToolkit/TimeSlider.swift

Toolkit/ArcGISToolkit/TimeSlider.swift

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+            var doMajorTickLabelsCollide = false
+            var firstMajorTickIndex = 0
+            let tickCount = tms.count
+            
+            // Calculate the largest number of ticks to allow between major ticks. This prevents scenarios where
+            // there are two major ticks placed undesirably close to the end.
+            let maxMajorTickInterval = Int(ceil(Double(tickCount / 2)))
+            
+            // Calculate the number of ticks between each major tick and the index of the first major tick
+            for i in majorTickInterval..<maxMajorTickInterval  {
```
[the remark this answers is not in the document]
njarecha: Added check to have at-least two ticks. and added condition here.

Toolkit/ArcGISToolkit/TimeSlider.swift
njarecha: See above comment.

njarecha: What would you recommend doing this here?

njarecha: Yes, fixed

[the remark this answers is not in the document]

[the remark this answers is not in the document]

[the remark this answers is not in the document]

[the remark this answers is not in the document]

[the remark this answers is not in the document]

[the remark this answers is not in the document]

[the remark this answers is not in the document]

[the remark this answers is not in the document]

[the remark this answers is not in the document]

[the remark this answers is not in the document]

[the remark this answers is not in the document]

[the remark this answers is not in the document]
njarecha: Changed to `var toSeconds`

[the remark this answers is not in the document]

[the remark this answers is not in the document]
njarecha: This has to be function because it takes parameters.

njarecha:
@esreli @philium I've addressed most of the code review comments but not all. I'll do so and let you know.

esreli:
Thanks @njarecha. Looking forward to a final skim after you've addressed the remaining comments. Cool stuff!

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+            
+            // Get the minimum and maximum allowable time step indexes. This is not necessarily the end of the time slider since
+            // the start and end times may be pinned.
+            let minTimeStepIndex = !isStartTimePinned ? 0 : startTimeStepIndex;
+            let maxTimeStepIndex = !isEndTimePinned ? ts.count - 1 : endTimeStepIndex;
+            
+            // Get the number of steps by which to move the current time.  If the number specified in the method call would move the current time extent
+            // beyond the valid range, clamp the number of steps to the maximum number that the extent can move in the specified direction.
+            var validTimeStepDelta = 0
+            if timeSteps > 0 {
```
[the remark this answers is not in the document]
njarecha: I feel this more cleaner without using min() and max() functions because it does not return the same value which is being checked in the condition.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+                validTimeStepDelta = startTimeStepIndex + timeSteps <= maxTimeStepIndex ? timeSteps : maxTimeStepIndex - startTimeStepIndex
+            }
+            else {
+                validTimeStepDelta = endTimeStepIndex + timeSteps >= minTimeStepIndex ? timeSteps : minTimeStepIndex - endTimeStepIndex
+            }
+            
+            // Get the new start and end time step indexes
+            let newStartTimeStepIndex = !isStartTimePinned && (validTimeStepDelta > 0 || startTimeStepIndex + validTimeStepDelta >= minTimeStepIndex) && ((endTimeStepIndex + validTimeStepDelta <= maxTimeStepIndex) || isEndTimePinned) ?
```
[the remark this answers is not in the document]

Toolkit/ArcGISToolkit/TimeSlider.swift
njarecha: `UIImage(named:` is not working. Hence, we are using bundle. Other toolkit components using too.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2539 @@
...
+        currentExtentEndTimeLabel.foregroundColor = currentExtentLabelColor.cgColor
+        currentExtentEndTimeLabel.alignmentMode = kCAAlignmentCenter
+        currentExtentEndTimeLabel.frame = CGRect.zero
+        currentExtentEndTimeLabel.contentsScale = UIScreen.main.scale
+        currentExtentEndTimeLabel.fontSize = currentExtentLabelFont.pointSize
+        layer.addSublayer(currentExtentEndTimeLabel)
+        
+        // Create the images
+        let bundle = Bundle(for: type(of: self))
+        playImage = UIImage(named: "Play", in: bundle, compatibleWith: nil)
```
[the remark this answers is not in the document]
njarecha: Change button size to 44 and images sizes to 44x44 - @1x, 88x88 - @2x, 132x132 - 3x (Will be added later on).

Toolkit/ArcGISToolkit/TimeSlider.swift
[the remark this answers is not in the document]
njarecha: I've created an internal issue to address for all toolkit components so this will be addressed later on.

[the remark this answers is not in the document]
njarecha: Remove commented code.

njarecha:
@esreli @philium I've addressed all PR review comments. Please re-review.

Examples/ArcGISToolkitExamples/TimeSliderExample.swift
```diff
@@ -0,0 +1,95 @@
...
+            timeSlider.leadingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.leadingAnchor, constant: margin).isActive = true
+            timeSlider.trailingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.trailingAnchor, constant: -margin).isActive = true
+        }
+        else {
+            timeSlider.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: margin).isActive = true
+            timeSlider.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: -margin).isActive = true
+        }
+        
+        // Add layer
+        let mapImageLayer = AGSArcGISMapImageLayer(url: URL(string: "https://rtc-100-3.esri.com/arcgis/rest/services/timeSupport/cycle_routes/MapServer")!)
```
philium: @njarecha Is this server publicly accessible? I'm getting the following error when selecting the Time Slider sample:

>A server with the specified hostname could not be found.

I am not getting that error with any of the other samples.

njarecha: No, the URL will change before merging this PR.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2609 @@
...
+            guard let strongSelf = self else {
+                return
+            }
+            
+            // Once, all layers are loaded,
+            // loop through all of them.
+            operationalLayers.forEach({ (layer) in
+                //
+                // The layer must be time aware, supports time filtering and time filtering is enabled.
+                if let timeAwareLayer = layer as? AGSTimeAware, timeAwareLayer.supportsTimeFiltering, timeAwareLayer.isTimeFilteringEnabled {
```
philium: @njarecha You might consider replacing this `if` statement with a `guard` statement.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2609 @@
...
+// MARK: - Time Slider Control
+
+public class TimeSlider: UIControl {
+    
+    // MARK: - Enumerations
+    
+    /**
+     Controls how labels on the slider are displayed.
+     */
+    public enum TimeSliderLabelMode {
```
philium: @njarecha This could be renamed to `LabelMode` now that it is a subtype of `TimeSlider`.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2609 @@
...
+    
+    private let minimumFrameWidth: CGFloat = 250.0
+    private let maximumThumbSize: CGFloat = 50.0
+    private var trackLayerSidePadding: CGFloat = 30.0
+    private let labelSidePadding: CGFloat = 10.0
+    private let labelPadding: CGFloat = 3.0
+    private let paddingBetweenLabels: CGFloat = 5.0
+    
+    private var tickMarkLabels = [CATextLayer]()
+    fileprivate var tickMarks: [TickMark]?
```
philium: @njarecha I would suggest making this non-optional.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2609 @@
...
+        b.addTarget(self, action: #selector(TimeSlider.backAction(_:)), for: .touchUpInside)
+        b.showsTouchWhenHighlighted = true
+        addSubview(b)
+        backButton = b
+        
+        // Refresh
+        refresh()
+    }
+    
+    private func refresh() {
```
philium: It seems to me that the contents of this function could be inlined into `layoutSubviews()`. And whenever a property changes that would affect layout, `setNeedsLayout()` should be called instead. That would prevent layout from happening more than once when multiple properties change.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2609 @@
...
+    
+    // MARK: Playback Properties
+    
+    /**
+     A Boolean value indicating whether playback buttons are visible or not. Default is true.
+     */
+    public var playbackButtonsVisible = true {
+        didSet {
+            layoutIfNeeded()
+            UIView.transition(with: self, duration: 0.5, options: .transitionCrossDissolve, animations: {
```
philium: @njarecha You might consider not implicitly animating changes to the time slider's properties. The user may not want them animated, or they may want to use a different duration, or they may want to specify a completion closure.

Toolkit/ArcGISToolkit/TimeSlider.swift

Toolkit/ArcGISToolkit/TimeSlider.swift

Toolkit/ArcGISToolkit/TimeSlider.swift

Toolkit/ArcGISToolkit/TimeSlider.swift
njarecha: Removed animation

njarecha:
@philium I've addressed some of the changes and using public URL. I'll address remaining comments later on.

philium:
@njarecha Thank you!

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2603 @@
...
+    private let maximumThumbSize: CGFloat = 50.0
+    private var trackLayerSidePadding: CGFloat = 30.0
+    private let labelSidePadding: CGFloat = 10.0
+    private let labelPadding: CGFloat = 3.0
+    private let paddingBetweenLabels: CGFloat = 5.0
+    
+    private var tickMarkLabels = [CATextLayer]()
+    fileprivate var tickMarks = [TickMark]()
+    
+    private var playPauseButton: UIButton?
```
philium: @njarecha You might consider making these buttons non-optional.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2603 @@
...
+            // there are two major ticks placed undesirably close to the end.
+            let maxMajorTickInterval = Int(ceil(Double(tickCount / 2)))
+            
+            // If maxMajorTickInterval is not greater than majorTickInterval
+            // then we can not do further calculation
+            if maxMajorTickInterval >= majorTickInterval {
+                //
+                // Calculate the number of ticks between each major tick and the index of the first major tick
+                for i in majorTickInterval..<maxMajorTickInterval  {
+                    let prospectiveInterv
```
philium: @njarecha While this works, you could let UIKit do the work for you:
1. Specify that the image is a template image in the Asset Catalogue. screen shot 2018-05-15 at 4 57 51 pm
2. Set the button's image using `setImage(_:for:)` rather than `setBackgroundImage(_:for:)`.
3. Set the button's `tintColor` to tint the image to the desired color.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2603 @@
...
+        // Create the images
+        let bundle = Bundle(for: type(of: self))
+        playImage = UIImage(named: "Play", in: bundle, compatibleWith: nil)
+        pauseImage = UIImage(named: "Pause", in: bundle, compatibleWith: nil)
+        forwardImage = UIImage(named: "Forward", in: bundle, compatibleWith: nil)
+        backImage = UIImage(named: "Back", in: bundle, compatibleWith: nil)
+        
+        // Add Play/Pause button
+        let p = UIButton(type: .custom)
+        p.setBackgroundImage(playImage, for: .normal)
```
philium: @njarecha You could save yourself some work by taking advantage of the fact that each state can have a separate image. For example, say that `normal` is the paused state (play image will be shown) and that `selected` is the playing state (pause image will be shown). You would only need to set the image for the two different states here and then update which image is shown by setting the button's state elsewhere.

Toolkit/ArcGISToolkit/TimeSlider.swift
philium: @esreli Specifying the bundle is necessary because UIImage.init(named:) looks in `Bundle.main`, which is the app's bundle.

Toolkit/ArcGISToolkit/TimeSlider.swift
philium: @njarecha It seems to me that `firstMajorTickIndex = tickCount / 2 - 1` would be equivalent. Similar on line `1644`.

philium: @njarecha Could you please make this a `struct` instead of a `class`?

Toolkit/ArcGISToolkit/TimeSlider.swift
njarecha: I want result as `Int` so this is fine.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2609 @@
...
+            currentExtentStartTimeLabel.fontSize = currentExtentLabelFont.pointSize
+            currentExtentEndTimeLabel.font = currentExtentLabelFont as CFTypeRef
+            currentExtentEndTimeLabel.fontSize = currentExtentLabelFont.pointSize
+        }
+    }
+    
+    /**
+     The date formatter used for full extent dates.
+     */
+    public var currentExtentDateFormatter: DateFormatter = {
```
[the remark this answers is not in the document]
njarecha: Remove formatter properties and added other public settable properties.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2609 @@
...
+    
+    // If set to True, it will show two thumbs, otherwise only one. Default is True.
+    fileprivate var isRangeEnabled: Bool = true {
+        didSet {
+            upperThumbLayer.isHidden = !isRangeEnabled
+            currentExtentEndTimeLabel.isHidden = !isRangeEnabled
+        }
+    }
+    
+    private var heightConstraint: NSLayoutConstraint?
```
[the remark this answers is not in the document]
njarecha: done

Toolkit/ArcGISToolkit/TimeSlider.swift
njarecha: done

Toolkit/ArcGISToolkit/TimeSlider.swift

Toolkit/ArcGISToolkit/TimeSlider.swift

Toolkit/ArcGISToolkit/TimeSlider.swift

Toolkit/ArcGISToolkit/TimeSlider.swift
njarecha: Are you saying `firstMajorTickIndex = tickCount / 2 - 1` and line `1644` will produce the same result?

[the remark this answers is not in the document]
njarecha: I'm not sure which line this comment is for. Could you please let me know. I can't see this comment in `Files changed` page.

njarecha: I'm not sure which line this comment is for. Could you please let me know. I can't see this comment in Files changed page.

njarecha:
@philium I've addressed remaining PR review comments. Please re-review.

Toolkit/ArcGISToolkit/TimeSlider.swift
philium: Let me break `Int(ceil(Double(tickCount / 2)))` down into steps:
```swift
let step1 = tickCount / 2
let step2 = Double(step1)
let step3 = ceil(step2)
let step4 = Int(step4)
```
Suppose there are `3` ticks. Let's look at each of the steps. 
```swift
let tickCount = 3 // 3 (Int)
let step1 = tickCount / 2 // step1 = 3 (Int) / 2 (Int) = 1 (Int)
let step2 = Double(step1) // step2 = Double(1) = 1.0 (Double)
let step3 = ceil(step2) // step3 = ceil(1.0) = 1.0 (Double)
let step4 = Int(step4) // step4 = Int(1.0) = 1 (Int)
```
Now suppose we cast `tickCount` to `Double` before doing anything else: `Int(ceil(Double(tickCount) / 2))`. Let me break that down into steps:
```swift
Int(ceil(Double(tickCount) / 2))
let step1 = Double(tickCount)
let step2 = step1 / 2
let step3 = ceil(step2)
let step4 = Int(step3)
```
Again, suppose there are `3`ticks:
```swift
let tickCount = 3 // 3 (Int)
let step1 = Double(tickCount) // step1 = Double(3) = 3.0 (Double)
let step2 = step1 / 2 // step2 = 3.0 (Double) / 2 (Double) = 1.5 (Double)
let step3 = ceil(step2) // step3 = ceil(1.5) = 2.0 (Double)
let step4 = Int(step3) // step4 = Int(2.0) = 2 (Int)
```
@njarecha Does that make sense?

Toolkit/ArcGISToolkit/TimeSlider.swift
philium: Yes. `tickCount` is an `Int`, so `tickCount / 2` will produce an `Int` using integer division. Therefore the cast to `Double`, the call to `trunc()` and the cast back to `Int` has no effect.

philium: @njarecha This is on line 2231: `class TickMark` -> `struct TickMark`.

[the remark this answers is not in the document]
philium: This is on line `2335`: `var label: CATextLayer?`.

Toolkit/ArcGISToolkit/TimeSlider.swift
njarecha: Yes, make sense now. Fixed

Toolkit/ArcGISToolkit/TimeSlider.swift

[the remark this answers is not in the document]
njarecha: Removed

njarecha: Objects of this class is being stored in the Array I expect them to be stored as reference. The `struct` will store them as copy and it does not work as expected. Hence, it should be `class`.

philium: Ah! I see what you mean. Okay, I suppose that's fine. 

It would be nice to make `TickMark` a value type. There would be a performance benefit and it would allow you to observe when `tickMarks` changes. For example, you could refactor updating tick mark labels out of `positionTickMarks()`:
```swift
fileprivate var tickMarks = [TickMark]() {
    didSet {
        guard tickMarks != oldValue else { return }
        updateTickMarkLabels()
    }
}
```

njarecha:
Closing this as new PR is created - https://github.com/Esri/arcgis-runtime-toolkit-ios/pull/43

PR #35, opened 2018-05-22
Refactors UnitsViewController

philium:
Refactors UnitsViewController to use modern APIs and be more streamlined.

### DIFF
--- a/Toolkit/ArcGISToolkit/MeasureToolbar.swift
+++ b/Toolkit/ArcGISToolkit/MeasureToolbar.swift
@@ -175,30 +175,39 @@ public class MeasureToolbar: UIView, AGSGeoViewTouchDelegate {
             bindToMapView(mapView: mapView)
         }
     }
-
-    public var linearUnits : [AGSLinearUnit]{
-        get { return unitsViewController.linearUnits }
-        set { unitsViewController.linearUnits = newValue }
+    
+    public var selectedLinearUnit: AGSLinearUnit = {
+        if NSLocale.current.usesMetricSystem {
+            return AGSLinearUnit.kilometers()
+        } else {
+            return AGSLinearUnit.miles()
+        }
+        }() {
+        didSet {
+            updateMeasurement()
+        }
     }
-    public var areaUnits : [AGSAreaUnit]{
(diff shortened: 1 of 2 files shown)


Toolkit/ArcGISToolkit/MeasureToolbar.swift
```diff
@@ -266,7 +278,9 @@ public class MeasureToolbar: UIView, AGSGeoViewTouchDelegate {
         segControl.setImage(measureFeatureImage, forSegmentAt: 2)
         segControlItem = UIBarButtonItem(customView: segControl)
         
-        resultView.buttonTapHandler = unitsButtonTap
```
philium: This was giving the result view a strong reference to the measure toolbar, thereby creating a retain cycle and preventing the toolbar from being deallocated.

Toolkit/ArcGISToolkit/MeasureToolbar.swift
```diff
@@ -297,30 +311,14 @@ public class MeasureToolbar: UIView, AGSGeoViewTouchDelegate {
         
         // notification
         
-        NotificationCenter.default.addObserver(forName: NSNotification.Name.AGSSketchEditorGeometryDidChange, object: lineSketchEditor, queue: nil, using: sketchDidChange)
```
philium: This (and the subsequent line) were also preventing the measure toolbar from being deallocated.

Toolkit/ArcGISToolkit/UnitsViewController.swift
```diff
@@ -125,155 +110,104 @@ public class UnitsViewController: TableViewController, UINavigationBarDelegate,
...
-        }
-    }
-    
-    public func position(for bar: UIBarPositioning) -> UIBarPosition {
-        return .topAttached
     }
     
     // MARK: TableView delegate/datasource methods
     
-    func tableView(_ tableView: UITableView, didSelectRowAtIndexPath indexPath: IndexPath) {
```
philium: This method had the wrong signature, so it was never getting called. Selecting a unit had not effect.

rolson:
Thanks @philium, @mhdostal can you please review?

PR #36, opened 2018-06-21
Ryan iphone x

rolson:
- fix compass
- update project to swift 4.1
- fixes for iPhone X


Toolkit/ArcGISToolkit/Compass.swift
```diff
@@ -67,20 +65,32 @@ public class Compass: UIImageView {
         
         // animate the compass visibility, if necessary
         animateCompass()
+        
+        // Add Compass as an observer of the mapView's rotation.
+        rotationObservation = mapView.observe(\.rotation, options: .new) { (mapView, change) in
```
philium: This closure needs to capture a `weak` reference to `self` in order to prevent a retain cycle.

Examples/ArcGISToolkitExamples.xcodeproj/project.pbxproj
```diff
@@ -178,7 +178,7 @@
 			isa = PBXProject;
 			attributes = {
 				LastSwiftUpdateCheck = 0800;
-				LastUpgradeCheck = 0820;
+				LastUpgradeCheck = 0930;
```
philium: Since you've performed the upgrade to Xcode 9.3, could you please check in the `IDEWorkspaceChecks.plist` file that Xcode creates automatically?

philium: Same with the ArcGISToolkit project.

PR #37, opened 2018-06-22
fixed doc, label placement bugs and added code to handle outstanding design issues

njarecha:
@philium @esreli I've made changes in the Time Slider for some of the outstanding design issues, added support for cases where data does not have `timeStepInterval`, fixed some of the label placement bugs and fixed ref doc. Can you please review this PR?

cc: @rolson

### DIFF
--- a/Toolkit/ArcGISToolkit/TimeSlider.swift
+++ b/Toolkit/ArcGISToolkit/TimeSlider.swift
@@ -868,7 +868,7 @@ public class TimeSlider: UIControl {
             scene = sceneView.scene
         }
         
-        // Set observeGeoView after setting map and scene
+        // Set observeGeoView after
         // setting map and scene
         self.observeGeoView = observeGeoView
         
--- a/Toolkit/ArcGISToolkit/TimeSlider.swift
+++ b/Toolkit/ArcGISToolkit/TimeSlider.swift
@@ -859,7 +859,6 @@ public class TimeSlider: UIControl {
         // Set geoview and it's properties on the slider
         // it will initialize required properties.
         self.geoView = geoView
-        self.observeGeoView = observeGeoView
         
         // Set map/scene
         if let mapView = geoView as? AGSMapView {
@@ -868,6 +867,10 @@ public class TimeSlider: UIControl {
         else if let sceneView = geoView as? AGSSceneView {
             scene = sceneView.scene
         }
+        
+        // Set observeGeoView after setting map and scene
+        // setting map and scene
+        self.observeGeoView = observeGeoView
         
         //
         // Loop through all time aware layers which are visible and are participating in time based filtering
--- a/Toolkit/ArcGISToolkit/TimeSlider.swift
+++ b/Toolkit/ArcGISToolkit/TimeSlider.swift
@@ -549,7 +549,16 @@ public class TimeSlider: UIControl {
     /**
      A boolean value indicating whether changes (change in layers and time extent) in geo view are being observed by the time slider.
      */
-    public var observeGeoView = false
+    public var observeGeoView = false {
+        didSet {
+            if observeGeoView {
+                addObservers()
+            }
+            else {
+                removeObservers()
+            }
+        }
+    }
     
     /**
      Time zone of the slider.
--- a/Toolkit/ArcGISToolkit/TimeSlider.swift
+++ b/Toolkit/ArcGISToolkit/TimeSlider.swift
@@ -869,11 +869,6 @@ public class TimeSlider: UIControl {
             scene = sceneView.scene
         }
         
-        // Start observing
-        if observeGeoView {
-            addObservers()
-        }
-        
         //
         // Loop through all time aware layers which are visible and are participating in time based filtering
         // and initialize slider's fullExtent, currentExtent and timeStepInterval properties
--- a/Toolkit/ArcGISToolkit/TimeSlider.swift
+++ b/Toolkit/ArcGISToolkit/TimeSlider.swift
@@ -542,12 +542,12 @@ public class TimeSlider: UIControl {
     // MARK: Other Properties
     
     /**
-     The geo view used to initialize the time slider properties.
+     The geoView used to initialize the time slider properties.
      */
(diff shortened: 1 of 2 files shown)


Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -185,7 +185,7 @@ public class TimeSlider: UIControl {
     }
     
     /**
-     A Boolean value that indicates whether the start time of the currentExtent can be
+     A boolean value that indicates whether the start time of the currentExtent can be
```
philium: Not sure what prompted this change, but Boolean is capitalized in the New Oxford American dictionary and in all of Apple's API documentation.

njarecha:
@mhdostal Can you please review this PR?

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -2410,49 +2475,48 @@ private struct DateRange : Sequence, IteratorProtocol {
 extension Date {
     //
     // Returns the amount of years from another date
-    func years(from date: Date) -> Int {
+    private func years(from date: Date) -> Int {
```
philium: Rather than marking each individual method as `private`/`fileprivate`, you could mark the entire extension as private:

```swift
private extension Date { … }
```

Toolkit/ArcGISToolkit/TimeSlider.swift
njarecha: Yes, you are right. I did not check the dictionary. Fixed.

Toolkit/ArcGISToolkit/TimeSlider.swift
njarecha: Marked extension `fileprivate`

njarecha:
@philium Please re-review.

njarecha:
@philium Can you please review my last commit?

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -869,6 +868,10 @@ public class TimeSlider: UIControl {
             scene = sceneView.scene
         }
         
+        // Set observeGeoView after setting map and scene
+        // setting map and scene
```
philium: It looks like this line is a partial copy of the previous line.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -542,11 +542,24 @@ public class TimeSlider: UIControl {
     // MARK: Other Properties
     
     /**
-     This will initialize the slider's fullExtent, currentExtent and timeStepInterval
-     properties based on time enabled layers in the geo view.
+     The geo view used to initialize the time slider properties.
```
mhdostal: "geo view" is usually one word in the doc:  `geoView`.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -542,11 +542,24 @@ public class TimeSlider: UIControl {
...
      */
     public private(set) var geoView: AGSGeoView?
     
+    /**
+     A Boolean value indicating whether changes (change in layers and time extent) in geo view are being observed by the time slider.
```
mhdostal: "geo view" -> `geoView`;
Also, say "are being observed", seems to indicate a passive operation by the user.  I think this should be "should be observed", because it's something the user can set if they want to.

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -832,20 +845,21 @@ public class TimeSlider: UIControl {
     
     /**
      This will look for all time aware layers which are visible and are participating in time based filtering
-     to initialize slider's fullExtent, currentExtent and timeStepInterval properties.
+     to initialize slider's fullExtent, currentExtent and timeStepInterval properties. Setting observeGeoView
```
mhdostal: It seems like we need to indicate in the doc that `observeGeoView` is a property.  Is that accomplished via `#observeGeoView`?

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -2138,6 +2180,32 @@ public class TimeSlider: UIControl {
         return dateFormatter.string(from: date)
     }
     
+    // Calculates time step interval based on provided time extent and time step count
+    private func calculateTimeStepInterval(for timeExtent: AGSTimeExtent, timeStepCount: Int) -> AGSTimeValue? {
+        if let startTime = timeExtent.startTime, let endTime = timeExtent.endTime {
+            //
+            // Calculate time step interval
+            // based on time step count
+            if timeStepCount > 0 {
```
mhdostal: Since you are dividing by `timeStepCount - 1` in the line below, should the check be for `timeStepCount > 1`?  To prevent divide-by-zero?

Or can it never be 1?

Toolkit/ArcGISToolkit/TimeSlider.swift
mhdostal: Nevermind, forgot we're not in the SDK.

Toolkit/ArcGISToolkit/TimeSlider.swift
njarecha: `timeStepCount` is going to be either `0` or `timeStepCount >= 2` (guard is in the `initializeTimeSteps` function from where this function is called) so the `divide-by-zero` won't happen. Do you still think I should change the condition?

Toolkit/ArcGISToolkit/TimeSlider.swift

Toolkit/ArcGISToolkit/TimeSlider.swift

njarecha:
@mhdostal Please re-review.

Toolkit/ArcGISToolkit/TimeSlider.swift
mhdostal: Ok, I see where the `guard` is.  I would still change the condition, as you're never sure if sometime later it will be called somewhere else.

PR #43, opened 2018-07-03
New Component: Time Slider

njarecha:


mhdostal:
@njarecha - this has all been reviewed already, correct?

njarecha:
@mhdostal Yes

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2727 @@
...
+    private let paddingBetweenLabels: CGFloat = 5.0
+    
+    private var tickMarkLabels = [CATextLayer]()
+    fileprivate var tickMarks = [TickMark]()
+    
+    private var playPauseButton = UIButton(type: .custom)
+    private var forwardButton = UIButton(type: .custom)
+    private var backButton = UIButton(type: .custom)
+    
+    private var playImage: UIImage?
```
philium: These image properties are only being used once and only inside of `setupUI()`. Could we get rid of them by either:
1. Moving them into `setupUI()`, or
1. Creating and setting the images simultaneously: `setImage(UIImage(named: …`

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2727 @@
...
+    private let maximumThumbSize: CGFloat = 50.0
+    private var trackLayerSidePadding: CGFloat = 30.0
+    private let labelSidePadding: CGFloat = 10.0
+    private let labelPadding: CGFloat = 3.0
+    private let paddingBetweenLabels: CGFloat = 5.0
+    
+    private var tickMarkLabels = [CATextLayer]()
+    fileprivate var tickMarks = [TickMark]()
+    
+    private var playPauseButton = UIButton(type: .custom)
```
philium: These button properties are never mutated after initialization. Could you make them constants (`let`)?

Toolkit/ArcGISToolkit/TimeSlider.swift
```diff
@@ -0,0 +1,2727 @@
...
+        
+        // Setup back button
+        backButton.setImage(backImage, for: .normal)
+        backButton.tintColor = playbackButtonsFillColor
+        backButton.addTarget(self, action: #selector(TimeSlider.backAction(_:)), for: .touchUpInside)
+        backButton.showsTouchWhenHighlighted = true
+        addSubview(backButton)
+        
+        // Refresh
+        setNeedsLayout()
```
philium: While not harmful, I don't think this is needed. The system should request that the view layout its subviews before it is displayed.

Toolkit/ArcGISToolkit/TimeSlider.swift
njarecha: Moved inside `setupUI()`

Toolkit/ArcGISToolkit/TimeSlider.swift

Toolkit/ArcGISToolkit/TimeSlider.swift
njarecha: Removed

njarecha:
@philium Thanks for the review. Addressed all the comments. Please re-review.

PR #44, opened 2018-10-10
Fix deprecated warnings

mhdostal:
This fixes two "deprecated" warnings when building against the v100.4 SDK.  The selection color property now lives on the `selectionProperties` property of `AGSMapView`.

The code to initialize the symbols with the selection color was moved to the `bindToMapView` method from the `sharedInitialization` method.  That is because at the time `sharedInitialization` is called, the `mapView` property hasn't been initialized yet, so the `mapView.selectionProperties.color` property will always be nil.

### DIFF
--- a/Toolkit/ArcGISToolkit/MeasureToolbar.swift
+++ b/Toolkit/ArcGISToolkit/MeasureToolbar.swift
@@ -318,7 +318,7 @@ public class MeasureToolbar: UIToolbar, AGSGeoViewTouchDelegate {
         
         // defaults for symbology
         selectionLineSymbol = lineSketchEditor.style.lineSymbol
-        selectionColor = lineSketchEditor.style.selectionColor
+        selectionColor = mapView?.selectionProperties.color
         let fillColor = (selectionColor ?? UIColor.cyan).withAlphaComponent(0.25)
         let sfs = AGSSimpleFillSymbol(style: .solid, color: fillColor, outline: selectionLineSymbol as? AGSSimpleLineSymbol)
         selectionFillSymbol = sfs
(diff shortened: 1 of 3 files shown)


Toolkit/ArcGISToolkit/MeasureToolbar.swift
```diff
@@ -174,8 +174,11 @@ public class MeasureToolbar: UIToolbar, AGSGeoViewTouchDelegate {
     // Exposed so that the symbology and selection colors can be customized.
     public private(set) var selectionLineSymbol : AGSSymbol?
     public private(set)var selectionFillSymbol : AGSSymbol?
-    public private(set) var selectionColor : UIColor?
-    
+    @available(iOS, deprecated, message: "Use `color` property exposed through `AGSGeoView#selectionProperties`")
```
philium: screen shot 2018-10-11 at 1 40 31 pm

Does the message get parsed for documentation? If not, I would suggest not using `AGSGeoView#selectionProperties` as it is a non-standard symbol name. The Swift symbol name is `AGSGeoView.selectionProperties`. You could use that, or you might consider using the key path:

screen shot 2018-10-11 at 1 42 09 pm

philium: You also might consider moving the property into an extension so that it doesn't clutter up the class implementation.

mhdostal: I added the `.` instead of `#`, but left the form of the message the same, since it matches the SDK API's message.

Thanks.

dg0yal:
Ideally this PR would have gone into a staging branch instead of master.
It would have then gone into master only when 100.4 SDK was released

mhdostal:
@dg0yal Yup, my bad.

PR #46, opened 2018-10-18
Mhd/release fixes

mhdostal:
Release updates for version #, Xcode project file path name and updated content insets for mapView in Measure Toolbar Examples.

### DIFF
--- a/Examples/ArcGISToolkitExamples/MeasureExample.swift
+++ b/Examples/ArcGISToolkitExamples/MeasureExample.swift
@@ -46,7 +46,7 @@ class MeasureExample: MapViewController{
         super.viewDidLayoutSubviews()
         
         // update content inset for mapview
-        mapView.contentInset = UIEdgeInsetsMake(0, 0, measureToolbar.frame.height, 0)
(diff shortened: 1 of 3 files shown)


Examples/ArcGISToolkitExamples/MeasureExample.swift
```diff
@@ -46,14 +46,7 @@ class MeasureExample: MapViewController{
...
         // update content inset for mapview
-        let tbHeight = measureToolbar.frame.height
-        
-        if #available(iOS 11.0, *) {
-            mapView.contentInset = UIEdgeInsetsMake(0, 0, view.safeAreaInsets.bottom + tbHeight, 0)
-        }
-        else{
-            mapView.contentInset = UIEdgeInsetsMake(0, 0, tbHeight, 0)
-        }
+        mapView.contentInset = UIEdgeInsetsMake(0, 0, measureToolbar.frame.height, 0)
```
philium: I would suggest switching to using the `UIEdgeInset` initializer `init(top:left:bottom:right:)`:
```suggestion
        mapView.contentInset = UIEdgeInsets(top: 0, left: 0, bottom: measureToolbar.frame.height, right: 0)
```

PR #50, opened 2019-04-09
Mhd/project updates

mhdostal:
Ran the recommended "Project Updates" in Xcode 10.2
Update Readme version numbers for Xcode, iOS SDK, Runtime SDK
Updated project version numbers to 100.5
Remove unreachable code from `MeasureToolbar.swift`

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,9 +14,9 @@ Toolkit components that will simplify your iOS app development with ArcGIS Runti
 * [TemplatePickerViewController](Documentation/TemplatePicker)
 
 ## Requirements
-* [ArcGIS Runtime SDK for iOS](https://developers.arcgis.com/en/ios/) 100.4.0 (or higher)
-* Xcode 9.3 (or higher)
-* Apple iOS SDK 10 (or higher)
+* [ArcGIS Runtime SDK for iOS](https://developers.arcgis.com/en/ios/) 100.5.0 (or higher)
+* Xcode 10.1 (or higher)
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -14,9 +14,9 @@ Toolkit components that will simplify your iOS app development with ArcGIS Runti
 * [TemplatePickerViewController](Documentation/TemplatePicker)
 
 ## Requirements
-* [ArcGIS Runtime SDK for iOS](https://developers.arcgis.com/en/ios/) 100.4.0 (or higher)
-* Xcode 9.3 (or higher)
-* Apple iOS SDK 10 (or higher)
+* [ArcGIS Runtime SDK for iOS](https://developers.arcgis.com/en/ios/) 100.5.0 (or higher)
+* Xcode 10.1 (or higher)
+* Apple iOS SDK 12.1 (or higher)
```
philium: Is this the minimum deployment target or base SDK? The minimum deployment target would now be iOS 11.0, correct?

mhdostal: That's the same number as in the Samples; It's the iOS SDK version that ships with Xcode 10.1, correct?

philium: > It's the iOS SDK version that ships with Xcode 10.1, correct?

Correct.

> That's the same number as in the Samples

So it is! I'll update the README to clarify that iOS 12.1 is the minimum base SDK and iOS 11.0 is the minimum deployment target.

mhdostal: Send me your changes and I'll duplicate them in the Toolkit.

PR #51, opened 2019-04-10
fix time slider issue when using in combination of play/pause and manual dragging

njarecha:
@mhdostal Please review.

The fix includes reported by you and another issue I found during testing.


mhdostal:
I tested the fix and it resolves the issue I was seeing.

rolson:
nice work 👍

PR #57, opened 2019-06-03
AR

mhdostal:
This is the initial implementation of AR functionality for the Toolkit.  The `mhd/AR_MergeBranch` is a dummy branch strictly for the PR.  The `AR` branch, going forward, will be the point of reference for all AR-related work and changes to that branch will be via PRs.  This PR is to review the initial work done and pushed directly to the `AR` branch.

There are two main files:

- ARExample.swift - the Toolkit Example App example for AR
- ArcGISARView.swift - a view combining ARKit with and `AGSSceneView` into an easy-to-consume object that exposes the full range of ARKit functionality and tight coordination with `AGSSceneView`

Specific things to look for/review:

- Code organization; is it readable, logically laid out, and understandable?
- Swift - does it use latest (4.x) Swift features and follow our Swift programming guidelines; is it good Swift code?

Note that the `CLLocationManager`, associated user authorization code and code to start/stop location and heading updates (all in `ArcGISARView.swift`) *might* be replaced by an `AGSLocationDataSource`; there is a pending design issue to make the necessary changes to that class in order to support what's necessary for AR (and other use cases).

Don't be afraid to be brutal as I want this to be outstanding open source code...


Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,682 @@
...
+        
+    }
+    
+    /*
+     *  locationManager:didUpdateHeading:
+     *
+     *  Discussion:
+     *    Invoked when a new heading is available.
+     */
+    @available(iOS 3.0, *)
```
rolson: any `available` tag less than iOS <= 11 we shouldn't need.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,682 @@
...
+    public func locationManager(_ manager: CLLocationManager, didUpdateHeading newHeading: CLHeading) {
```
rolson: i think you can go ahead and remove any empty and optional delegate method

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,682 @@
...
+    }
+    
+    // MARK: Errors
+    
+    /// Error used when ARKit is not supported on the current device
+    ///
+    /// - Returns: error stating ARKit not supported
+    fileprivate class func notSupportedError() -> NSError {
+        let userInfo = [NSLocalizedDescriptionKey : "The device does not support ARKit functionality."]
+        return NSError(domain: AGSErrorDomain, code: 0, userInfo: userInfo)
```
rolson: is there an `AGSErrorCode` value that works for this? `AGSErrorCodeCommonIllegalState`?

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,682 @@
...
+// MARK: - ARSessionDelegate
+
+extension ArcGISARView: ARSessionDelegate {
+    
+    /**
+     This is called when a new frame has been updated.
+     
+     @param session The session being run.
+     @param frame The frame that has been updated.
+     */
```
rolson: can we change these to the swift style?

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,682 @@
...
+            let locationPoint = AGSPoint(x: location.coordinate.longitude,
+                                         y: location.coordinate.latitude,
+                                         z: location.altitude,
+                                         spatialReference: .wgs84())
+            let camera = AGSCamera(location: locationPoint, heading: 0.0, pitch: 0.0, roll: 0.0)
+            initialTransformationMatrix = camera.transformationMatrix
+            sceneView.setViewpointCamera(camera)
+            
+            finalizeStart()
+            print("didUpdateLocations - initialLocation...")
```
rolson: did you want/mean to leave print statements in?

Examples/ArcGISToolkitExamples/ARExample.swift
```diff
@@ -0,0 +1,76 @@
...
+    
+    override open func viewDidAppear(_ animated: Bool) {
+        arView.startTracking()
+    }
+    
+    override open func viewDidDisappear(_ animated: Bool) {
+        arView.stopTracking()
+    }
+    
+    private func scene() -> AGSScene {
```
rolson: you can make this a var (calculated property), or a lazy var (lazy calculated property) if you only ever want it created once.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,682 @@
...
+            self,
+            selector: #selector(self.orientationChanged(notification:)),
+            name: UIDevice.orientationDidChangeNotification,
+            object: nil
+        )
+        
+        //
+        // reset frameCount and start timer to capture frame rate
+        frameCount = 0
+        frameCountTimer = Timer.scheduledTimer(withTimeInterval: 1.0, repeats: true, block: { [weak self] (timer) in
```
philium: I was just about to submit my review when I happened to notice that Apple provided the code to do this without a `Timer` in the Platforms State of the Union:

IMG_2416

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,682 @@
...
+    private var initialTransformationMatrix = AGSTransformationMatrix()
+    
+    /// Whether ARKit supported on this device?
+    private var isSupported = false
+    
+    /// Whether the client has been notfiied of start/failure
+    private var notifiedStartOrFailure = false
+    
+    /// These are used when calculating framerate
+    var frameCount:Int = 0
```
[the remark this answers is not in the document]
philium: I should have just offered a suggestion:
```suggestion
    var frameCount: Int = 0
```

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
mhdostal: Does `AGSErrorCodeCommonPropertyNotSupported` work?

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
mhdostal: Removed.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
philium: How about `ARErrorCodeUnsupportedConfiguration`?

rolson: seems fine

mhdostal:
> I left a bunch of comments and suggestions. Please take a look and let me know what you think.
> 
> The SIMD API got a big overhaul in Swift 5. Have you considered migrating to Swift 5 and using the new SIMD types?

I don't think we're ready to move to Swift 5 yet.  It will require Xcode 10.2 and we only require 10.1.  Maybe at 100.7

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,682 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import UIKit
+import ARKit
+import ArcGIS
+
+public class ArcGISARView: UIView {
```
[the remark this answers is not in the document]
mhdostal: Hadn't thought about subclassing.  Would `open` be a better choice here?  Since this is open source, I'm not sure if there's a use case for subclassing...

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
mhdostal: I'll give that a try, thanks!

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,682 @@
...
+    
+    /// Whether the client has been notfiied of start/failure
+    private var notifiedStartOrFailure = false
+    
+    /// These are used when calculating framerate
+    var frameCount:Int = 0
+    var frameCountTimer: Timer?
+    
+    // A quaternion used to compensate for the pitch beeing 90 degrees on ARKit; used to calculate the current device transformation for each frame
+    let compensationQuat:simd_quatf = simd_quaternion(Float(sin(45 / (180 / Float.pi))), 0, 0, Float(cos(45 / (180 / Float.pi))))
```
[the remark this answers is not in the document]
mhdostal: Changed to sims_quatd; that was a leftover from the original Beta.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,682 @@
...
+
+import UIKit
+import ARKit
+import ArcGIS
+
+public class ArcGISARView: UIView {
+    
+    // MARK: public properties
+    
+    /// The view used to display the ARKit camera image and 3D SceneKit content
```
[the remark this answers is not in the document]
mhdostal: How do you do code formatting?  is it just `ARKit`?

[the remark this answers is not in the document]
philium: Yes, by enclosing it in back ticks:

```suggestion
    /// The view used to display the `ARKit` camera image and 3D `SceneKit` content.
```

Before | After
------ | -----
image | image

Examples/ArcGISToolkitExamples/ARExample.swift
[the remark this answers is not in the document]
mhdostal: Moved to property

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
mhdostal: Removed them.

Examples/ArcGISToolkitExamples/ARExample.swift
mhdostal: changed to property.

mhdostal:
@philium Thanks for all the great feedback.  I made the changes you suggested, please re-review.

Examples/ArcGISToolkitExamples/ARExample.swift
```diff
@@ -0,0 +1,71 @@
...
+import UIKit
+import ARKit
+import ArcGISToolkit
+import ArcGIS
+
+open class ARExample: UIViewController {
+    
+    public let arView = ArcGISARView(frame: CGRect.zero)
+    
+    private let scene: AGSScene = {
```
philium: Now that I see this implemented as a property, I regret suggesting it. There's a downside to the property approach that I missed: what happens if the user assigns their own scene to the scene view? This scene is no longer the correct scene and will not be deallocated until the view is deallocated. Since there's no way to prevent user's from changing the scene (as long as `arView` is public), I think using a method was the right approach, just with a better name.

Edit: I originally commented this in the wrong place.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,527 @@
...
+    }
+    
+    required public init?(coder aDecoder: NSCoder) {
+        super.init(coder: aDecoder)
+        sharedInitialization()
+    }
+    
+    /// Initializer used to denote whether to display the live camera image.
+    ///
+    /// - Parameter renderVideoFeed: whether to dispaly the live camera image.
```
philium: I would suggest starting parameter descriptions with an uppercase letter as well. Also, there's a typo:

```suggestion
    /// - Parameter renderVideoFeed: Whether to display the live camera image.
```

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,527 @@
...
+        
+        // Set intitial orientationQuat.
+        orientationChanged(notification: nil)
+    }
+    
+    // MARK: Public
+    
+    /// Determines the map point for the given screen point.
+    ///
+    /// - Parameter screenPoint: the point in screen coordinates.
```
```suggestion
    /// - Parameter screenPoint: The point in screen coordinates.
```

Examples/ArcGISToolkitExamples/ARExample.swift
mhdostal: Back to a method:  `private func makeStreetsScene() -> AGSScene`

I think that works better; it may be that we'll have several different scenes the user can choose from, so having a method with `makeXXX` works for me.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,527 @@
...
+        arSCNView.isHidden = !renderVideoFeed
+        
+        // Run the ARSession.
+        arSCNView.session.run(arConfiguration, options:.resetTracking)
+        didStartOrFailWithError(nil)
+    }
+
+    /// Adds subView to superView with appropriate constraints.
+    ///
+    /// - Parameter subview: the subView to add.
```
```suggestion
    /// - Parameter subview: The subView to add.
```

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,527 @@
...
+    }
+    
+    /// Start the locationManager with authorized access.
+    fileprivate func startWithAccessAuthorized() {
+        startUpdatingLocationAndHeading()
+    }
+    
+    /// Potential notification to the user of an error starting device tracking.
+    ///
+    /// - Parameter error: error that ocurred when starting tracking.
```
```suggestion
    /// - Parameter error: The error that occurred when starting tracking.
```

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,527 @@
...
+    /// Handle a change in authorization status to "authorized".
+    fileprivate func handleAuthStatusChangedAccessAuthorized() {
+        // Auth status changed to authorized; now that we have authorization - start updating location and heading.
+        didStartOrFailWithError(nil)
+        startUpdatingLocationAndHeading()
+    }
+    
+    /// Called when device orientation changes.
+    ///
+    /// - Parameter notification: the notification.
```
```suggestion
    /// - Parameter notification: The notification.
```

Examples/ArcGISToolkitExamples/ARExample.swift
```diff
@@ -0,0 +1,72 @@
...
+// limitations under the License.
+
+import UIKit
+import ARKit
+import ArcGISToolkit
+import ArcGIS
+
+open class ARExample: UIViewController {
+    
+    public let arView = ArcGISARView(frame: CGRect.zero)
```
philium:
```suggestion
    public let arView = ArcGISARView(frame: .zero)
```

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,527 @@
...
+    // We implement SCNSceneRendererDelegate methods, but will use `scnSceneRendererDelegate` to forward them to clients.
+    weak open var scnSceneRendererDelegate: SCNSceneRendererDelegate?
+
+    // MARK: private properties
+    
+    /// Whether to display the camera image or not.
+    private var renderVideoFeed = true
+    
+    /// Used to determine the device location when originCamera is not set.
+    private lazy var locationManager: CLLocationManager = {
```
philium: `self` really should be captured differently here:
```suggestion
    private lazy var locationManager: CLLocationManager = { [unowned self] in
```

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,527 @@
...
+        lm.desiredAccuracy = kCLLocationAccuracyBest
+        lm.delegate = self
+        return lm
+    }()
+    
+    /// Initial location from locationManager.
+    private var initialLocation: CLLocation?
+    
+    /// Current horizontal accuracy of the device.
+    private var horizontalAccuracy: CLLocationAccuracy = .greatestFiniteMagnitude;
```
```suggestion
    private var horizontalAccuracy: CLLocationAccuracy = .greatestFiniteMagnitude
```

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,527 @@
...
+    }
+    
+    /// Called when device orientation changes.
+    ///
+    /// - Parameter notification: the notification.
+    @objc func orientationChanged(notification: Notification?) {
+        // Handle rotation here.
+        switch UIApplication.shared.statusBarOrientation {
+        case .landscapeLeft:
+            orientationQuat = simd_quaternion(0, 0, 1.0, 0);
```
```suggestion
            orientationQuat = simd_quaternion(0, 0, 1.0, 0)
```

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,527 @@
...
+        case .landscapeRight:
+            orientationQuat = simd_quaternion(0, 0, 0, 1.0);
```
```suggestion
            orientationQuat = simd_quaternion(0, 0, 0, 1.0)
```

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,527 @@
...
+        case .portrait:
+            orientationQuat = simd_quaternion(0, 0, sqrt(0.5), sqrt(0.5));
```
```suggestion
            orientationQuat = simd_quaternion(0, 0, sqrt(0.5), sqrt(0.5))
```

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,527 @@
...
+        case .portraitUpsideDown:
+            orientationQuat = simd_quaternion(0, 0, -sqrt(0.5), sqrt(0.5));
```
```suggestion
            orientationQuat = simd_quaternion(0, 0, -sqrt(0.5), sqrt(0.5))
```

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,527 @@
...
+    }
+    
+    // MARK: Errors
+    
+    /// Error used when `ARKit` is not supported on the current device.
+    ///
+    /// - Returns: Error stating `ARKit` not supported.
+    fileprivate class func notSupportedError() -> NSError {
+        let userInfo = [NSLocalizedDescriptionKey : "The device does not support ARKit functionality."]
+        return NSError(domain: AGSErrorDomain, code: ARError.unsupportedConfiguration.rawValue, userInfo: userInfo)
```
philium: I missed this one. If you're going to use an `ARError` (which is fine), it should be in that domain:
```suggestion
        return NSError(domain: ARErrorDomain, code: ARError.unsupportedConfiguration.rawValue, userInfo: userInfo)
```

mhdostal:
> I made another round of suggestions (take them or leave them), but one crucial change I see is how `self` is captured when `locationManager` is created.

I'll go as many rounds as needed!  :-)

mhdostal:
@philium Pushed the latest set of changes.  One remaining question I had about this:

> I just want to confirm that you don't want to allow subclassing (since you used `public` instead of `open`). Is that correct?

Hadn't thought about subclassing. Would `open` be a better choice here? Since this is open source, I'm not sure if there's a use case for subclassing. Thoughts?

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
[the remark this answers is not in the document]
rolson: could probably allow subclassing later, but we can't dis-allow it later. I would suggest dis-allow it now and open it up later if necessary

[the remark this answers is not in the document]
mhdostal: So dis-allowing it now means leaving it as `public`, correct?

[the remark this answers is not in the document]
philium: Correct.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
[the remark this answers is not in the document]
mhdostal: Switched to calculating the frameRate in `willRender` and using the `time` argument.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,682 @@
...
+    }
+    
+    // MARK: Public
+    
+    /// Determines the map point for the given screen point
+    ///
+    /// - Parameter screenPoint: the point in screen coordinates
+    /// - Returns: the map point corresponding to screenPoint
+    public func arScreenToLocation(screenPoint: AGSPoint) -> AGSPoint {
+        return AGSPoint(x: 0.0, y: 0.0, spatialReference: nil)
```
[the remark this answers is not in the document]
mhdostal: The working of this is designed to correspond to the `AGSMapView`'s `-(AGSPoint*)screenToLocation:(CGPoint)screen` method.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,682 @@
...
+        // reset frameCount and start timer to capture frame rate
+        frameCount = 0
+        frameCountTimer = Timer.scheduledTimer(withTimeInterval: 1.0, repeats: true, block: { [weak self] (timer) in
+            print("Frame rate = \(String(reflecting: self?.frameCount))")
+            self?.frameCount = 0
+        })
+    }
+
+    /// Suspends device tracking
+    public func stopTracking() {
```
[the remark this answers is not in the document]
mhdostal: Done.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,682 @@
...
+        //
+        // run the ARSession
+        arSCNView.session.run(arConfiguration, options:.resetTracking)
+        didStartOrFailWithError(nil)
+    }
+
+    /// Adds subView to superView with appropriate constraints
+    ///
+    /// - Parameter subview: the subView to add
+    fileprivate func addSubviewWithConstraints(_ subview: UIView) {
```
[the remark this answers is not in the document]
mhdostal: Is there a preferred way to do this?  I thought Autolayout was the way to go...

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,682 @@
...
+    /// - Returns: error stating plist information is missing
+    fileprivate class func missingPListKeyError() -> NSError{
+        let userInfo = [NSLocalizedDescriptionKey : "You must specify a location usage description key (NSLocationWhenInUseUsageDescription or NSLocationAlwaysUsageDescription) in your plist."]
+        return NSError(domain: kCLErrorDomain, code: CLError.Code.denied.rawValue, userInfo: userInfo)
+    }
+}
+
+// MARK: - ARSessionDelegate
+
+extension ArcGISARView: ARSessionDelegate {
```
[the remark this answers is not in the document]
mhdostal: Is there a reason for them to be private?

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,682 @@
...
+    }
+
+    /// The viewpoint camera used to set the initial view of the sceneView instead of the devices GPS location via the locationManager
+    public var originCamera: AGSCamera?
+    
+    /// The translation factor used to support a table top AR experience
+    public var translationTransformationFactor: Double = 1.0
+    
+    // We implement ARSessionDelegate methods, but will use `sessionDelegate` to forward them to clients
+    weak open var sessionDelegate: ARSessionDelegate?
```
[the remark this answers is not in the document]
mhdostal: Good points.  I will look at this further...

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,682 @@
...
+            subview.trailingAnchor.constraint(equalTo: self.trailingAnchor),
+            subview.topAnchor.constraint(equalTo: self.topAnchor),
+            subview.bottomAnchor.constraint(equalTo: self.bottomAnchor)
+            ])
+    }
+    
+    /// Start the locationManager with undetermined access
+    fileprivate func startWithAccessNotDetermined() {
+        if (Bundle.main.object(forInfoDictionaryKey: "NSLocationWhenInUseUsageDescription") != nil) {
+            locationManager.requestWhenInUseAuthorization()
```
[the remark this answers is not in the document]
mhdostal: We may be switching all this code out in favor of a true `LocationDataSource`, but I'll re-examine this in the meantime.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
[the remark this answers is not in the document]
philium: > Is there a preferred way to do this?

Not really. The preferred way is probably whatever is simplest/easiest.

> I thought Autolayout was the way to go

Using the autoresizing mask is still using Auto Layout when `translatesAutoresizingMaskIntoConstraints` is enabled. The system takes the value of the autoresizing mask and generates the appropriate constraints automagically.

To be clear: There isn't a right way and a wrong way. What you have is just fine and either approach should produce the exact same constraints. The `autoresizingMask` route might just be a bit simpler.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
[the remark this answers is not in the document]
philium: Good question. The answer is that it depends. There are at least two considerations:

1. Do you want the outside world to know that you conform to the protocol in question and which methods you implement?
2. Because the outside world knows about them, there is the potential that they could be called by someone other than the object of which you are the delegate.

I generally view conformances like this as implementation details and try to keep them internal. But the indirection does take extra work.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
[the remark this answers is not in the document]
philium: > does this make more sense?

I don't think so since `toLocation` is a parameter label, but the parameter is the `from`, not the `to`. AppKit had a similar API design challenge with methods like `UIWindow`'s  `- (NSRect)convertRectToScreen:(NSRect)rect`. If I remember correctly, the Swift interface generator originally turned that into `convert(toScreen:)`, but they changed it to `convertToScreen(_:)` since the parameter label didn't make sense.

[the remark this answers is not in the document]
mhdostal: Ok, thanks.  Back to `arScreenToLocation`

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,682 @@
...
+    
+    /// Resets the device tracking using a spacial anchor
+    ///
+    /// - Returns: reset operation success or failure
+    public func resetUsingSpatialAnchor() -> Bool {
+        return false
+    }
+    
+    /// Starts device tracking
+    public func startTracking() {
```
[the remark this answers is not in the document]
mhdostal: The method itself cannot fail, but starting the tracking can.  This follows the pattern in the `AGSLocationDataSource`.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,682 @@
...
+            case .restricted, .denied:
+                startWithAccessDenied()
+            case .authorizedAlways, .authorizedWhenInUse:
+                startWithAccessAuthorized()
+            }
+        }
+        
+        //
+        // we need to know when the device orientation changes in order to update the Camera transformation
+        UIDevice.current.beginGeneratingDeviceOrientationNotifications()
```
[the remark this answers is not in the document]
mhdostal: `endGeneratingDeviceOrientationNotifications()` is called in `stopTracking()`.

How would you go about determining when the orientation of the device changes?

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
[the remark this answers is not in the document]
mhdostal: Ok, thanks.  I'm going to leave it as is because, during testing, I've needed to modify the size of one or both of the subViews and having the constraints spelled out makes that easier.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,682 @@
...
+    fileprivate func startWithAccessAuthorized() {
+        startUpdatingLocationAndHeading()
+    }
+    
+    /// Potential notification to the user of an error starting device tracking
+    ///
+    /// - Parameter error: error that ocurred when starting tracking
+    fileprivate func didStartOrFailWithError(_ error: Error?) {
+        if !notifiedStartOrFailure, let error = error {
+            // TODO: present error to user...
```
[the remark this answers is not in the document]
mhdostal: The current thought is that all of the location and heading stuff will be removed from here and replaced with an `AGSLocationDataSource`.  That will simplify a lot of the code and it has it's own mechanism for notifying clients of errors.  If it's OK with you, I'd like to leave this as is (with maybe a note that it will be changing).

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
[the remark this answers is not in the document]
mhdostal: Is `UITraitEnvironment` the way to go for this stuff, along with `UITraitCollection`?  I haven't read much about them...

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,682 @@
...
+        // need to start location manager updates
+        startUpdatingLocationAndHeading()
+    }
+    
+    /// Called when device orientation changes
+    ///
+    /// - Parameter notification: the notification
+    @objc func orientationChanged(notification: Notification?) {
+        // handle rotation here
+        switch UIApplication.shared.statusBarOrientation {
```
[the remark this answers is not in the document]
mhdostal: Made the change.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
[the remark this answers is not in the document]
philium: My experience has been with video, not with AR. In the past I've observed the bounds/frame size changes in `layoutSubviews()` to determine the orientation of a view, but it looks like you really need the device orientation. `ARKit` looks be an entirely different thing altogether. What you're doing may be the correct way; I just don't know. If what you're doing is working for you, then forget I said anything.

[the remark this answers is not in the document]
philium: > Is UITraitEnvironment the way to go for this stuff, along with UITraitCollection?
I don't think so. `UITraitCollection` doesn't contain device or interface orientation information.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,531 @@
...
+        didStartOrFailWithError(nil)
+        startUpdatingLocationAndHeading()
+    }
+    
+    /// Called when device orientation changes.
+    ///
+    /// - Parameter notification: The notification.
+    @objc func orientationChanged(notification: Notification?) {
+        // Handle rotation here.
+        switch UIApplication.shared.statusBarOrientation {
```
philium: A couple things I just noticed:
- You're subscribing to `UIDevice.orientationDidChangeNotification`, but you're checking `UIApplication.statusBarOrientation`. Perhaps you mean to check `UIDevice.orientation` or instead subscribe to `UIApplication. didChangeStatusBarOrientationNotification`?
- `UIApplication.statusBarOrientation` is deprecated.

mhdostal: Fixed; using UIDevice.current.orientation, but I will look into the other things I mentioned.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,682 @@
...
+    
+    /// Initialization code shared between all initializers
+    private func sharedInitialization(){
+        //
+        // ARKit initialization
+        isSupported = ARWorldTrackingConfiguration.isSupported
+        
+        addSubviewWithConstraints(arSCNView)
+        arSCNView.session.delegate = self
+        (arSCNView as SCNSceneRenderer).delegate = self
```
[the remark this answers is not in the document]
mhdostal: Conforming to `ARSCNViewDelegate`.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,682 @@
...
+            break
+        }
+    }
+    
+    // MARK: Errors
+    
+    /// Error used when ARKit is not supported on the current device
+    ///
+    /// - Returns: error stating ARKit not supported
+    fileprivate class func notSupportedError() -> NSError {
```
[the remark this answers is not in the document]
mhdostal: I updated how the errors are created.  If/when we switch to using `AGSLocationDataSource`, the error stuff will go away.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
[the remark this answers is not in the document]
mhdostal: I'm OK with having them public.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,682 @@
...
+        DispatchQueue.main.async {
+            // Present an alert describing the error.
+            let alertController = UIAlertController(title: "Could not start tracking.", message: errorMessage, preferredStyle: .alert)
+            let restartAction = UIAlertAction(title: "Restart Tracking", style: .default) { _ in
+                alertController.dismiss(animated: true, completion: nil)
+                self.startTracking()
+            }
+            alertController.addAction(restartAction)
+            
+            guard let rootController = UIApplication.shared.keyWindow?.rootViewController else { return }
```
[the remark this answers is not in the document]
mhdostal: Moved to the ARExample VC

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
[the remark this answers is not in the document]
mhdostal: it's hard to know what clients will or won't want to implement.  We need to one or two of them, so I'm not sure how to make sure clients get those methods called, other than to implement them all and pass through.

As for new delegate methods, that shouldn't happen except for major iOS releases and even then the code is open source, so clients can add whatever they need.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,682 @@
...
+            errorWithInfo.localizedRecoverySuggestion
+        ]
+        
+        // Remove optional error messages.
+        let errorMessage = messages.compactMap({ $0 }).joined(separator: "\n")
+        
+        DispatchQueue.main.async {
+            // Present an alert describing the error.
+            let alertController = UIAlertController(title: "Could not start tracking.", message: errorMessage, preferredStyle: .alert)
+            let restartAction = UIAlertAction(title: "Restart Tracking", style: .default) { _ in
```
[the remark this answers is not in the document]
mhdostal: No option to cancel.  If they cancelled, the app wouldn't function as it's dependent on ARKit up and running.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,682 @@
...
+     and tracking state will switch to limited with reason relocalizing. If successful, the
+     session's tracking state will return to normal. Because relocalization depends on
+     the user's location, it can run indefinitely. Apps that wish to give up on relocalization
+     may call run with `ARSessionRunOptionResetTracking` at any time.
+     @param session The session to relocalize.
+     @return Return YES to begin relocalizing.
+     */
+    @available(iOS 11.3, *)
+    public func sessionShouldAttemptRelocalization(_ session: ARSession) -> Bool {
+        if let result = sessionDelegate?.sessionShouldAttemptRelocalization?(session) {
```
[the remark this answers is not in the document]
mhdostal: Done.

mhdostal:
@philium I have addressed all your comments.  Take a specific look at the Error stuff, as that's probably the biggest change I'm most unsure about.

Thanks again for all your feedback!

Examples/ArcGISToolkitExamples/ARExample.swift
```diff
@@ -0,0 +1,102 @@
...
+        ]
+        
+        // Remove optional error messages.
+        let errorMessage = messages.compactMap({ $0 }).joined(separator: "\n")
+        
+        DispatchQueue.main.async { [unowned self] in
+            // Present an alert describing the error.
+            let alertController = UIAlertController(title: "Could not start tracking.", message: errorMessage, preferredStyle: .alert)
+            let restartAction = UIAlertAction(title: "Restart Tracking", style: .default) { _ in
+                alertController.dismiss(animated: true)
```
philium: `UIAlertController` automatically dismisses itself when an action is chosen. Were you finding you needed to call this explicitly?

Examples/ArcGISToolkitExamples/ARExample.swift
```diff
@@ -0,0 +1,102 @@
...
+        let messages = [
+            errorWithInfo.localizedDescription,
+            errorWithInfo.localizedFailureReason,
+            errorWithInfo.localizedRecoverySuggestion
+        ]
+        
+        // Remove optional error messages.
+        let errorMessage = messages.compactMap({ $0 }).joined(separator: "\n")
+        
+        DispatchQueue.main.async { [unowned self] in
```
philium: There's no guarantee that `self` will be around when this block is executed on the main queue. Therefore you need to weakly capture `self` here.

To prevent having to include multiple references to `self`, I usually extract the contents of a closure like this into a method, and then just delegate to that method include the closure. Something like this:

```swift
DispatchQueue.main.async { [weak self] in self?.showError(withMessage: errorMessage) }
```

Examples/ArcGISToolkitExamples/ARExample.swift
```diff
@@ -0,0 +1,102 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import UIKit
+import ARKit
+import ArcGISToolkit
+import ArcGIS
+
+open class ARExample: UIViewController {
```
philium: Since this class isn't used outside of this module (the app), it doesn't need to be `open` (or `public`). It would be best to just use the default (`internal`). Same for all of its members.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,538 @@
...
+    ///
+    /// - Parameter notification: The notification.
+    @objc func orientationChanged(notification: Notification?) {
+        // Handle rotation here.
+        orientationQuat = simd_quatd(statusBarOrientation: UIDevice.current.orientation)
+    }
+}
+
+// MARK: - ARSessionDelegate
+extension ArcGISARView: ARSessionDelegate {
```
philium: Are you responding to any of these methods yourself anymore? If not, maybe you could just assign `arSCNViewDelegate` to `arSCNView.delegate` directly? Or better yet, let users of `ArcGISARView` do it themselves? That way you wouldn't have to do all of this manual message forwarding.

Examples/ArcGISToolkitExamples/ARExample.swift
mhdostal: Leftover code from long ago...

Examples/ArcGISToolkitExamples/ARExample.swift
mhdostal: Got it, thanks.  Since there were only two references to self, I did not break the code out into a method.

Examples/ArcGISToolkitExamples/ARExample.swift
mhdostal: Done.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
mhdostal: Removed the ARSessionDelegate stuff as it's not used currently.

mhdostal:
@philium Made the changes we talked about; please re-review.  Thanks!

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,512 @@
...
+            horizontalAccuracy = location.horizontalAccuracy
+        }
+        
+    }
+    
+    public func locationManager(_ manager: CLLocationManager, didFailWithError error: Error) {
+        didStartOrFailWithError(error)
+    }
+    
+    public  func locationManager(_ manager: CLLocationManager, didChangeAuthorization status: CLAuthorizationStatus) {
```
```suggestion
    public func locationManager(_ manager: CLLocationManager, didChangeAuthorization status: CLAuthorizationStatus) {
```

mhdostal:
@rolson Phil has approved the PR; please review.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,512 @@
...
+            return "Access to the device location is denied."
+        case .missingPListKey:
+            return "You must specify a location usage description key (NSLocationWhenInUseUsageDescription or NSLocationAlwaysUsageDescription) in your plist."
+        }
+    }
+}
+
+public class ArcGISARView: UIView {
+    
+    enum clError {
```
rolson: Since this is a typename, shouldn't it be `CLError`?

rolson: Actually, this isn't a proper "Error" object, so it probably should be named something more like `CLErrorType`?

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,512 @@
...
+        case missingPListKey
+    }
+
+    // MARK: public properties
+    
+    /// The view used to display the `ARKit` camera image and 3D `SceneKit` content.
+    public lazy private(set) var arSCNView = ARSCNView(frame: .zero)
+    
+    /// The view used to display ArcGIS 3D content.
+    public lazy private(set) var sceneView = AGSSceneView(frame: .zero)
```
rolson: I don't see a reason these should be lazy. They are used right away in sharedInitialization, which is called from the init. Couldn't it just be constants like so:

```swift
public let sceneView = AGSSceneView(frame: .zero)
```

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,512 @@
...
+    
+    /// The world tracking information used by `ARKit`.
+    public var arConfiguration: ARConfiguration = {
+        let config = ARWorldTrackingConfiguration()
+        config.worldAlignment = .gravityAndHeading
+        return config
```
rolson: indentation seems off here maybe

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,512 @@
...
+    // MARK: public properties
+    
+    /// The view used to display the `ARKit` camera image and 3D `SceneKit` content.
+    public lazy private(set) var arSCNView = ARSCNView(frame: .zero)
+    
+    /// The view used to display ArcGIS 3D content.
+    public lazy private(set) var sceneView = AGSSceneView(frame: .zero)
+    
+    /// The world tracking information used by `ARKit`.
+    public var arConfiguration: ARConfiguration = {
```
rolson: is this missing the `lazy` keyword?

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,512 @@
...
+    /// Whether `ARKit` is supported on this device.
+    private var isSupported = {
+        return ARWorldTrackingConfiguration.isSupported
+    }()
+    
+    /// Whether the client has been notfiied of start/failure.
+    private var notifiedStartOrFailure = false
+    
+    /// Used when calculating framerate.
+    private var lastUpdateTime: TimeInterval = 0
```
rolson: for a `TimeInterval` property, I'm not sure the name `lastUpdateTIme` makes sense. Is it more "timeSinceLastUpdate`?

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,512 @@
...
+    
+    // A quaternion used to compensate for the pitch beeing 90 degrees on `ARKit`; used to calculate the current device transformation for each frame.
+    let compensationQuat:simd_quatd = simd_quatd(ix: (sin(45 / (180 / .pi))), iy: 0, iz: 0, r: (cos(45 / (180 / .pi))))
```
rolson: need a space after the colon

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,512 @@
...
+    
+    /// The quaternion used to represent the device orientation; used to calculate the current device transformation for each frame; defaults to landcape-left.
+    var orientationQuat:simd_quatd = simd_quatd(statusBarOrientation: .landscapeLeft)
```
rolson: space after colon

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,512 @@
...
+        
+        // Add sceneView to view and setup constraints.
+        addSubviewWithConstraints(sceneView)
+
+        // Make our sceneView's background transparent, no atmosphereEffect.
+        sceneView.isBackgroundTransparent = true
+        sceneView.atmosphereEffect = .none
+        
+        // Tell the sceneView we will be calling `renderFrame()` manually.
+        sceneView.isManualRendering = true
```
rolson: do you want to do this even if not showing the video feed from the camera?

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,512 @@
...
+            case .notDetermined:
+                startWithAccessNotDetermined()
+            case .restricted, .denied:
+                startWithAccessDenied()
+            case .authorizedAlways, .authorizedWhenInUse:
+                startWithAccessAuthorized()
+            }
+        }
+        
+        //TODO:  reloook at the mechanism by which we're getting notified of orientation changed events...
```
rolson: relook

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,512 @@
...
+        // We need to know when the device orientation changes in order to update the Camera transformation.
+        UIDevice.current.beginGeneratingDeviceOrientationNotifications()
+        NotificationCenter.default.addObserver(
```
rolson: In some cases, this won't happen until after finalizeStart is called. `finalizeStart` looks like it makes callbacks, which if at that point the user does something to stopTracking, then the observer will get removed from the notification center before it's added. This used to be a crash (not sure if it still is, but I think so), however, at the least what could happen is that the observer is then added after it's removed, so it won't get properly removed.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,512 @@
...
+        startUpdatingLocationAndHeading()
+    }
+    
+    /// Potential notification to the user of an error starting device tracking.
+    ///
+    /// - Parameter error: The error that occurred when starting tracking.
+    fileprivate func didStartOrFailWithError(_ error: Error?) {
+        if !notifiedStartOrFailure, let error = error {
+            // TODO: present error to user...
+            print("didStartOrFailWithError: \(String(reflecting:error))")
```
rolson: do you want to leave print statements in?

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,512 @@
...
+    fileprivate func handleAuthStatusChangedAccessDenied() {
+        // auth status changed to denied
+        stopUpdatingLocationAndHeading()
+        
+        // We were waiting for user prompt to come back, so notify.
+        didStartOrFailWithError(ArcGISARView.clError.accessDenied)
+    }
+    
+    /// Handle a change in authorization status to "authorized".
+    fileprivate func handleAuthStatusChangedAccessAuthorized() {
```
rolson: I can see why you want to re-use the location datasource, because it does some of this code for you. However - that is really written to power the location display. It sets it's own location manager up, and it might be setup differently than you need it setup for AR. 

It might be best to just move this out into an extension or another object just for AR?

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -0,0 +1,512 @@
...
+        transformationMatrix = initialTransformationMatrix.addTransformation(transformationMatrix)
+        
+        let camera = AGSCamera(transformationMatrix: transformationMatrix)
+        sceneView.setViewpointCamera(camera)
+        
+        sceneView.renderFrame()
+
+        // Calculate frame rate.
+        let frametime = time - lastUpdateTime
+        print("Frame rate = \(String(reflecting: Int((1.0 / frametime).rounded())))")
```
rolson: this should result in thousands of print statements to the console. Not sure that is ideal

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
mhdostal: The Re-indent feature did not make a change.  It does look funny, but I think that's because it's a compute property with a `didSet` observer.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
mhdostal: Done.  Made both `let`.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
mhdostal: `lastUpdateTime` is used in `willRenderScene`.  The `time` argument of the [willRenderScene]() method is defined as: _The current system time, in seconds. If your custom rendering involves animation, use this parameter to compute your own animation state._

`lastUpdateTime` is set to that value at the end of the method.  So it's not an interval, but an actual time.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
mhdostal: I had it `lazy` at first, but got this compile error: `Lazy properties must not have observers`.  So it can't be `lazy`.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
mhdostal: I deleted the `didStartOrFailWithError:` one out, but just commented out the `frameRate` one, as that will be useful during development.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
mhdostal: Commented it out.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
rolson: ok

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
mhdostal: I modified the error stuff a bit; changed the enum name to `CoreLocationError` and it now inherits from Swift.Error.  There already is a `CLError` in the CoreLocation stuff.

mhdostal: See a link from @philium on Swift errors [here](https://nshipster.com/swift-foundation-error-protocols/)

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
mhdostal: I'm not sure.  I am going to enter an issue in our repo so the team can talk about it.  That OK?

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
mhdostal: I removed that comment; I took another look at it after Phil's review and I'm comfortable with how it works.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
rolson: ok

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
mhdostal: I really would like to get rid of all that code as it was copied from the `AGSCLLocationDataSource`.  There's a big push to use the `LocationDatasource` and the design has already been changed..  The `AGSCLLocationDataSource` has a property for the `CLLocationManager` (`@property (nonatomic, strong, readonly) CLLocationManager *locationManager;`) so I think we'll be OK.

rolson: I'm starting to feel like the LocationDatasource object to use. if we change the location manager that the map is using, that will have affects on how the map is getting it's locations, which may or may not be desired.

Just my thought right now is CLLocationManager is the object to go to for getting locations, the LocationDatasource...

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
mhdostal: Added a boolean tracking whether the observer was added or not.  Also now using the more specific `removeObserver(_:name:object:)` method per the Apple doc.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
rolson: Although it seems that ship has sailed, at least for the moment. So let's ignore my comment for now

mhdostal:
@rolson @philium I'm done with the last (final?) batch of updates, please re-review.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -330,7 +331,6 @@ public class ArcGISARView: UIView {
     fileprivate func didStartOrFailWithError(_ error: Error?) {
```
rolson: if this takes a `ArcGISARView.CoreLocationError` instead of an `Error`, then calling code can/should just look like this:

```swift
didStartOrFailWithError(.missingPListKey)
```

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -294,7 +295,7 @@ public class ArcGISARView: UIView {
             locationManager.requestAlwaysAuthorization()
         }
         else{
-            didStartOrFailWithError(ArcGISARView.clError.missingPListKey)
+            didStartOrFailWithError(ArcGISARView.CoreLocationError.missingPListKey)
```
rolson: within ArcGISARView you don't need to qualify `ArcGISARView.CoreLocationError.missingPListKey`. You can just write `CoreLocationError.missingPListKey`

mhdostal: Yes, thank you.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
mhdostal: It takes an `Error`, so `CoreLocationError.missingPListKey` works.

mhdostal:
@rolson Done again, please review.

mhdostal:
`mhd/AR_MergeBranch` was a dummy for the PR.  Approved code will still reside in the AR branch, but going forward, all work needs to be done in a separate branch and the PR reviewed prior to going into the AR branch.

PR #60, opened 2019-06-25
Mhd/fov

mhdostal:
This adds code to set the `SceneView`'s field of view from the ARKit camera intrinsics.

### DIFF
--- a/Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
+++ b/Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
@@ -17,6 +17,9 @@ import ARKit
 import ArcGIS
 
 extension AGSDeviceOrientation {
+    /// Allows creation of an `AGSDeviceOrientation` from a `UIDeviceOrientation`.
+    ///
+    /// - Parameter statusBarOrientation: The `UIDeviceOrientation` to create the `AGSDeviceOrientation` from.
     init?(statusBarOrientation: UIDeviceOrientation) {
         switch statusBarOrientation {
         case .landscapeLeft:
--- a/Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
+++ b/Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
@@ -20,19 +20,18 @@ extension AGSDeviceOrientation {
     /// Allows creation of an `AGSDeviceOrientation` from a `UIDeviceOrientation`.
     ///
     /// - Parameter statusBarOrientation: The `UIDeviceOrientation` to create the `AGSDeviceOrientation` from.
-    init?(statusBarOrientation: UIDeviceOrientation) {
+    init(statusBarOrientation: UIDeviceOrientation) {
         switch statusBarOrientation {
         case .landscapeLeft:
-            self.init(rawValue: AGSDeviceOrientation.landscapeRight.rawValue)
+            self = .landscapeRight
         case .landscapeRight:
-            self.init(rawValue: AGSDeviceOrientation.landscapeLeft.rawValue)
+            self = .landscapeLeft
         case .portrait:
-            self.init(rawValue: AGSDeviceOrientation.portrait.rawValue)
+            self = .portrait
         case .portraitUpsideDown:
-            self.init(rawValue: AGSDeviceOrientation.reversePortrait.rawValue)
+            self = .reversePortrait
         default:
-            // default to landscapeLeft
-            self.init(rawValue: AGSDeviceOrientation.landscapeLeft.rawValue)
+            self = .landscapeRight
         }
     }
 }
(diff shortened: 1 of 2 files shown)


Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -407,6 +428,19 @@ extension ArcGISARView: SCNSceneRendererDelegate {
...
+        if let camera = arSCNView.session.currentFrame?.camera {
+            let intrinsics = camera.intrinsics
+            let imageResolution = camera.imageResolution
+            sceneView.setFieldOfViewFromLensIntrinsicsWithXFocalLength(intrinsics[0][0],
+                                                                       yFocalLength: intrinsics[1][1],
+                                                                       xPrincipal: intrinsics[2][0],
+                                                                       yPrincipal: intrinsics[2][1],
+                                                                       xImageSize: Float(imageResolution.width),
+                                                                       yImageSize: Float(imageResolution.height),
+                                                                       deviceOrientation: AGSDeviceOrientation.init(statusBarOrientation: UIDevice.current.orientation) ?? .landscapeRight)
```
```suggestion
                                                                       deviceOrientation: AGSDeviceOrientation(statusBarOrientation: UIDevice.current.orientation) ?? .landscapeRight)
```
or
```suggestion
                                                                       deviceOrientation: .init(statusBarOrientation: UIDevice.current.orientation) ?? .landscapeRight)
```

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -16,6 +16,27 @@ import UIKit
 import ARKit
 import ArcGIS
 
+extension AGSDeviceOrientation {
+    /// Allows creation of an `AGSDeviceOrientation` from a `UIDeviceOrientation`.
+    ///
+    /// - Parameter statusBarOrientation: The `UIDeviceOrientation` to create the `AGSDeviceOrientation` from.
+    init?(statusBarOrientation: UIDeviceOrientation) {
```
philium: This initializer is failable, but it doesn't actually ever fail (it never returns `nil`). Perhaps it either shouldn't be failable or the `default` case should be removed?

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -16,6 +16,27 @@ import UIKit
...
+        switch statusBarOrientation {
+        case .landscapeLeft:
+            self.init(rawValue: AGSDeviceOrientation.landscapeRight.rawValue)
```
philium: Should this be `landscapeLeft`?

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -16,6 +16,27 @@ import UIKit
...
+        case .landscapeRight:
+            self.init(rawValue: AGSDeviceOrientation.landscapeLeft.rawValue)
```
philium: Should this be `landscapeRight`?

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
mhdostal: The SceneView API has them reversed from the UIDeviceOrientation values.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
mhdostal: The SceneView API has them reversed from the UIDeviceOrientation values.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
mhdostal: The initializer can fail because the code in each case is fail-able:

`self.init(rawValue: AGSDeviceOrientation.landscapeRight.rawValue)`  //<<== this can fail

I thought the same thing (that it wouldn't be fail-able), but it's a compile error to not make it optional.

mhdostal:
@philium Made the suggested changes and commented on the rest, please re-review.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
philium: My apologies. I didn't look closely enough and assumed you were force unwrapping the initialized value.

I would suggest not using `init(rawValue:)` to make the initializer non-failable:

```swift
switch statusBarOrientation {
case .landscapeLeft:
    self = .landscapeRight
case .landscapeRight:
    self = .landscapeLeft
case .portrait:
    self = .portrait
case .portraitUpsideDown:
    self = .reversePortrait
default:
    self = .landscapeLeft
}
```

I commented that last time around, but must have hit the wrong button.

mhdostal: That much cleaner, thanks!

mhdostal:
@rolson Can you review?

PR #62, opened 2019-07-18
Mhd/final ar screen to location

mhdostal:
https://devtopia.esri.com/runtime/queen-annes-revenge/issues/101

Implements both of the `setInitialTransform` methods and the `arScreenToLocation` method.

### DIFF
--- a/Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
+++ b/Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
@@ -201,7 +201,7 @@ public class ArcGISARView: UIView {
         return true
     }
     
-    /// Sets the initial transformation used to offset the originCamera.  The initial transformation is based on an AR point determined via existing plan hit detection from `screenPoint`.  If an AR point cannot be determined, this method will return `false`.
+    /// Sets the initial transformation used to offset the originCamera.  The initial transformation is based on an AR point determined via existing plane hit detection from `screenPoint`.  If an AR point cannot be determined, this method will return `false`.
     ///
     /// - Parameter screenPoint: The screen point to determine the `initialTransformation` from.
     /// - Returns: Whether setting the `initialTransformation` succeeded or failed.
--- a/Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
+++ b/Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
@@ -172,7 +172,7 @@ public class ArcGISARView: UIView {
     
     /// Determines the map point for the given screen point.
     ///
-    /// - Parameter toLocation: The point in screen coordinates.
+    /// - Parameter screenPoint: The point in screen coordinates.
     /// - Returns: The map point corresponding to screenPoint.
     public func arScreenToLocation(screenPoint: CGPoint) -> AGSPoint? {
         // Use the `internalHitTest` method to get the matrix of `screenPoint`.
--- a/Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
+++ b/Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
@@ -24,11 +24,7 @@ public class ArcGISARView: UIView {
     public let arSCNView = ARSCNView(frame: .zero)
     
     /// The initial transformation used for a table top experience.  Defaults to the Identity Matrix.
-    public var initialTransformation: AGSTransformationMatrix {
-        get {
-            return _initialTransformation
-        }
-    }
+    public var initialTransformation: AGSTransformationMatrix = AGSTransformationMatrix.identity
     
     /// Denotes whether tracking location and angles has started.
     public private(set) var isTracking: Bool = false
@@ -101,9 +97,6 @@ public class ArcGISARView: UIView {
     private let deviceSupportsARKit: Bool = {
         return ARWorldTrackingConfiguration.isSupported
     }()
-    
-    /// Internal readwrite initialTransformation property
-    private var _initialTransformation = AGSTransformationMatrix.identity
 
     // MARK: Initializers
     
@@ -197,7 +190,7 @@ public class ArcGISARView: UIView {
     /// - Parameter initialTransformation: The initial transformation for originCamera offset.
     /// - Returns: Whether setting the `initialTransformation` succeeded or failed.
     public func setInitialTransformation(_ initialTransformation: AGSTransformationMatrix) -> Bool {
-        _initialTransformation = initialTransformation
+        self.initialTransformation = initialTransformation
         return true
     }
     
@@ -210,7 +203,7 @@ public class ArcGISARView: UIView {
         guard let matrix = internalHitTest(screenPoint: screenPoint) else { return false }
         
         // Set the `initialTransformation` as the AGSTransformationMatrix.identity - hit test matrix.
-        _initialTransformation = AGSTransformationMatrix.identity.subtractTransformation(matrix)
+        initialTransformation = AGSTransformationMatrix.identity.subtractTransformation(matrix)
 
         return true
     }
--- a/Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
+++ b/Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
@@ -24,7 +24,11 @@ public class ArcGISARView: UIView {
     public let arSCNView = ARSCNView(frame: .zero)
     
     /// The initial transformation used for a table top experience.  Defaults to the Identity Matrix.
-    public private(set) var initialTransformation = AGSTransformationMatrix.identity
+    public var initialTransformation: AGSTransformationMatrix {
+        get {
+            return _initialTransformation
+        }
+    }
     
     /// Denotes whether tracking location and angles has started.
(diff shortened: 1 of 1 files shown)


Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -245,14 +269,40 @@ public class ArcGISARView: UIView {
...
+        // Get the worldTransform from the first result; if there's no worldTransform, return nil.
+        guard let worldTransform = results.first?.worldTransform else { return nil }
+        
+        // Create our hit test matrix based on the worldTransform location.
+        let hitTestMatrix = AGSTransformationMatrix(quaternionX: 0.0,
+                                                    quaternionY: 0.0,
+                                                    quaternionZ: 0.0,
+                                                    quaternionW: 1.0,
+                                                    translationX: Double(worldTransform.columns.3.x),
+                                                    translationY: Double(-worldTransform.columns.3.z),
```
pgruenler: Just wanted to check that z to Y and y to Z is on purpose.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -24,7 +24,11 @@ public class ArcGISARView: UIView {
     public let arSCNView = ARSCNView(frame: .zero)
     
     /// The initial transformation used for a table top experience.  Defaults to the Identity Matrix.
-    public private(set) var initialTransformation = AGSTransformationMatrix.identity
```
philium: This is the natural way of doing this sort of thing. I suspect you moved away from it because of this?

```swift
public func setInitialTransformation(initialTransformation: AGSTransformationMatrix) -> Bool {
    _initialTransformation = initialTransformation
    return true
}
```

If so, the solution is to add `self`:

```swift
self.initialTransformation = initialTransformation
```

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -165,10 +174,18 @@ public class ArcGISARView: UIView {
     ///
     /// - Parameter toLocation: The point in screen coordinates.
```
```suggestion
    /// - Parameter screenPoint: The point in screen coordinates.
```

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -180,17 +197,24 @@ public class ArcGISARView: UIView {
     /// - Parameter initialTransformation: The initial transformation for originCamera offset.
     /// - Returns: Whether setting the `initialTransformation` succeeded or failed.
     public func setInitialTransformation(initialTransformation: AGSTransformationMatrix) -> Bool {
```
```suggestion
    public func setInitialTransformation(_ initialTransformation: AGSTransformationMatrix) -> Bool {
```

philium: I'm guessing the return value is for future use? Normally I would suggest that failure operation should throw, but that might be a bit heavy handed in this case. I would suggest that you annotate the method with `@discardableResult` as a convenience for callers who don't care whether the operation succeeded. Same with the other `setInitialTransformation` methods.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -180,17 +197,24 @@ public class ArcGISARView: UIView {
...
-        fatalError("setInitialTransformation(initialTransformation:) has not been implemented")
+        _initialTransformation = initialTransformation
+        return true
     }
     
     /// Sets the initial transformation used to offset the originCamera.  The initial transformation is based on an AR point determined via existing plan hit detection from `screenPoint`.  If an AR point cannot be determined, this method will return `false`.
```
philium: Should that be "plane"?

```suggestion
    /// Sets the initial transformation used to offset the originCamera.  The initial transformation is based on an AR point determined via existing plane hit detection from `screenPoint`.  If an AR point cannot be determined, this method will return `false`.
```

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -180,17 +197,24 @@ public class ArcGISARView: UIView {
...
     ///
     /// - Parameter screenPoint: The screen point to determine the `initialTransformation` from.
     /// - Returns: Whether setting the `initialTransformation` succeeded or failed.
-    public func setInitialTransformation(screenPoint: CGPoint) -> Bool {
-        fatalError("setInitialTransformationPlacingOriginOnPlane(screenPoint:) has not been implemented")
-    }
+    public func setInitialTransformation(_ screenPoint: CGPoint) -> Bool {
```
philium: This looks a bit confusing at the call site:

```swift
view.setInitialTransformation(CGPoint(...))
```

The method name makes it sound like we're setting a transformation, but the parameter type is a point. Perhaps a parameter label is in order?

```suggestion
    public func setInitialTransformation(using screenPoint: CGPoint) -> Bool {
```

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -245,14 +269,40 @@ public class ArcGISARView: UIView {
             subview.bottomAnchor.constraint(equalTo: self.bottomAnchor)
             ])
     }
+    
+    /// Internal method to perform a hit test operation to get the transformation matrix representing the corresponding real-world point for `screenPoint`.
+    ///
+    /// - Parameter screenPoint: screenPoint: The screen point to determine the real world transformation matrix from.
```
```suggestion
    /// - Parameter screenPoint: The screen point to determine the real world transformation matrix from.
```

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
mhdostal: Yes, that is on purpose.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
mhdostal: The `setInitialTransformation(...)` methods were interfering with the `initialTransformation` property when I had it the original way (`public private(set) var...`).  So I made the property a get-only and have a backing private variable.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
mhdostal: I think the other method, which *can* return `false` needs the non-discardable result.  The reason this one has a return value of bool is that it is an "overloaded" method, so in the design the signature was the same except for the argument.

I was going back-and-forth debating whether to just remove the Bool return value, since it's always going to be true (and it seems to compile fine with it removed).

Thoughts?

philium: I see what you mean. Part of me thinks that the return value isn't necessary for any of the functions since the user can check the value of `initialTransformation` to see if it has changed. But that's up to you. I don't see a big problem with different functions in the same family having different return types.

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
philium: I'm not sure the `setInitialTransformation` methods themselves were the cause of the problem. I was able to build using the old `initialTransformation` property by updating all of the references from `_initialTransformation` to `initialTransformation`, with the exception of the above mentioned method needing to disambiguate `initialTransformation` (local variable) from `initialTransformation` (instance variable:

```swift
public func setInitialTransformation(_ initialTransformation: AGSTransformationMatrix) -> Bool {
    self.initialTransformation = initialTransformation
    return true
}
```

What you have is functionally equivalent, so it's fine. I just wanted you to know that what you had before should work (with a slight tweak).

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
mhdostal: The `Bool` return value is necessary for the method that takes a screenPoint as the hit test the operation is based on may return `nil`.  I'm going to leave the `Bool` return for the other `setInitialTransformation` method as that is what the design states, but I will add the `@discardableResult` keyword.

mhdostal:
@philium I made the changes, please re-review (GitHub seems to working well enough now...)

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -24,7 +24,7 @@ public class ArcGISARView: UIView {
     public let arSCNView = ARSCNView(frame: .zero)
     
     /// The initial transformation used for a table top experience.  Defaults to the Identity Matrix.
-    public private(set) var initialTransformation = AGSTransformationMatrix.identity
+    public var initialTransformation: AGSTransformationMatrix = AGSTransformationMatrix.identity
```
philium: This was `private(set) before. I assume you still want that? Also, you're specifying the type twice. While not wrong, it is unnecessary. You could omit one of them. Either:

```suggestion
    public private(set) var initialTransformation = AGSTransformationMatrix.identity
```

or

```suggestion
    public private(set) var initialTransformation: AGSTransformationMatrix = .identity
```

mhdostal:
@pgruenler Please re-review.

PR #63, opened 2019-07-23
JobManager clean up

nixta:
- Remove the odd `$0.1` reference to a dictionary tuple's value and replace with a more descriptive `$0.value`.
- Rename some non-descriptive variable names.
- Tidy up the coding style around opening block statements to be more consistent.
- Refactor some iterations to be more consistent across the class.

### DIFF
--- a/Toolkit/ArcGISToolkit/JobManager.swift
+++ b/Toolkit/ArcGISToolkit/JobManager.swift
@@ -234,6 +234,8 @@ public class JobManager: NSObject {
     
     /// Resume all paused and not-started jobs.
     public func resumeAllPausedJobs(statusHandler: @escaping JobStatusHandler, completion: @escaping JobCompletionHandler) {
+        // Note that a job is also paused if it is created from JSON. So any jobs that have been stored in
+        // UserDefaults and reloaded will be in the paused state.
         keyedJobs.filter({ $0.value.status == .paused || $0.value.status == .notStarted }).forEach {
             $0.value.start(statusHandler: statusHandler, completion:completion)
         }
--- a/Toolkit/ArcGISToolkit/JobManager.swift
+++ b/Toolkit/ArcGISToolkit/JobManager.swift
@@ -46,15 +46,17 @@ public class JobManager: NSObject {
         return _jobManagerSharedInstance
     }
     
+    /// The JobManager ID, provided during initialization.
     public let jobManagerID: String
     
-    // Flag to signify that we shouldn't write to defaults
-    // Maybe we are currently reading from the defaults so it's pointless to write to them.
-    // Or maybe we are waiting until a group of modifications are made before writing to the defaults.
+    /// Flag to signify that we shouldn't write to User Defaults.
+    ///
+    /// Used internally when reading stored `AGSJob`s from the User Defaults during init().
     private var suppressSaveToUserDefaults = false
     
     private var kvoContext = 0
     
+    /// A dictionary of Unique IDs and `AGSJob`s that the `JobManager` is managing.
     public private(set) var keyedJobs = [String: AGSJob]() {
         willSet {
             // Need `self` because of a Swift bug.
@@ -63,13 +65,14 @@ public class JobManager: NSObject {
         didSet {
             keyedJobs.values.forEach { observeJobStatus(job: $0) }
 
-            // If there was a change, then re-store the serialized jobs in UserDefaults
+            // If there was a change, then re-store the serialized AGSJobs in UserDefaults
             if keyedJobs != oldValue {
                 saveJobsToUserDefaults()
             }
         }
     }
 
+    /// A convenience accessor to the `AGSJob`s that the `JobManager` is managing.
     public var jobs: [AGSJob] {
         return Array(keyedJobs.values)
     }
@@ -79,6 +82,8 @@ public class JobManager: NSObject {
     }
     
     /// Create a JobManager with an ID.
+    ///
+    /// - Parameter jobManagerID: An arbitrary identifier for this JobManager.
     public required init(jobManagerID: String) {
         self.jobManagerID = jobManagerID
         super.init()
@@ -93,8 +98,7 @@ public class JobManager: NSObject {
         return keyedJobs.compactMapValues { try? $0.toJSON() }
     }
     
-    // observing job status code
-    
+    // Observing job status code
     private func observeJobStatus(job: AGSJob) {
         job.addObserver(self, forKeyPath: #keyPath(AGSJob.status), context: &kvoContext)
     }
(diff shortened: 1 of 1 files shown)


nixta:
I had an inkling that might be the case.

Toolkit/ArcGISToolkit/JobManager.swift
```diff
@@ -167,49 +168,43 @@ public class JobManager: NSObject {
...
     /// Clears the finished Jobs from the Job manager.
-    public func clearFinishedJobs(){
-        
+    public func clearFinishedJobs() {
         suppressSaveToUserDefaults = true
-        for (jobUniqueID, job) in keyedJobs{
-            if job.status == .failed || job.status == .succeeded{
-                keyedJobs.removeValue(forKey: jobUniqueID)
-            }
+        keyedJobs.filter({ $0.value.status == .failed || $0.value.status == .succeeded }).forEach {
```
philium: This potentially mutates `keyedJobs` multiple times. Instead, filter out the jobs you don't want and assign that back to `keyedJobs`:

```swift
public func clearFinishedJobs() {
    keyedJobs = keyedJobs.filter { (_, value) in
        let status = value.status
        return !(status == .failed || status == .succeeded)
    }
}
```

mhdostal:
@nixta - this should be merged into the `update6` branch; once we're ready to release and everything has been tested, it will go into master from there.

Toolkit/ArcGISToolkit/JobManager.swift
```diff
@@ -167,49 +168,43 @@ public class JobManager: NSObject {
...
+                if error != nil {
                     completedWithoutErrors = false
                 }
                 group.leave()
             }
             cancelGroup.children.append(cancellable)
         }
         
-        group.notify(queue: DispatchQueue.main){
+        group.notify(queue: DispatchQueue.main) {
```
philium:
```suggestion
        group.notify(queue: .main) {
```

Toolkit/ArcGISToolkit/JobManager.swift
```diff
@@ -222,26 +217,27 @@ public class JobManager: NSObject {
...
-        if self.jobs.count > 0{
-            self.checkStatusForAllJobs{ completedWithoutErrors in
-                if completedWithoutErrors{
-                    completionHandler(.newData)
-                }
-                else{
-                    completionHandler(.failed)
-                }
-            }
+        guard jobs.count > 0 else {
```
philium:
```suggestion
        guard !jobs.isEmpty else {
```

Although I believe performing work in the `else` clause goes against the spirit of `guard` because forward progress is made either way:

```swift
if !jobs.isEmpty {
    checkStatusForAllJobs { completedWithoutErrors in
        if completedWithoutErrors {
            completionHandler(.newData)
        } else {
            completionHandler(.failed)
        }
    }
} else {
    completionHandler(.noData)
}
```

Toolkit/ArcGISToolkit/JobManager.swift
```diff
@@ -222,26 +217,27 @@ public class JobManager: NSObject {
...
 
     
     /// Resume all paused and not-started jobs.
-    public func resumeAllPausedJobs(statusHandler: @escaping JobStatusHandler, completion: @escaping JobCompletionHandler){
-        keyedJobs.filter{ $0.1.status == .paused || $0.1.status == .notStarted}.forEach{
-            $0.1.start(statusHandler: statusHandler, completion:completion)
+    public func resumeAllPausedJobs(statusHandler: @escaping JobStatusHandler, completion: @escaping JobCompletionHandler) {
+        // Note that a job is also paused if it is created from JSON. So any jobs that have been stored in
+        // UserDefaults and reloaded will be in the paused state.
+        keyedJobs.filter({ $0.value.status == .paused || $0.value.status == .notStarted }).forEach {
```
philium: Lazily filtering would prevent looping twice:

```suggestion
        keyedJobs.lazy.filter({ $0.value.status == .paused || $0.value.status == .notStarted }).forEach {
```

Toolkit/ArcGISToolkit/JobManager.swift
```diff
@@ -76,38 +77,39 @@ public class JobManager: NSObject {
         jobs.forEach { unObserveJobStatus(job: $0) }
         
         // set new jobs array
-        jobs = keyedJobs.map{ $0.1 }
+        jobs = keyedJobs.map{ $0.value }
```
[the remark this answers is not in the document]
nixta: > self.keyedJobs.values.forEach(unObserveJobStatus(job:))

Do you know why I need the `self` here? It [seems it was considered a bug](https://stackoverflow.com/questions/29343654/attempting-to-store-to-property-within-its-own-willset), but it also seems it was fixed a few releases ago. Regression?

Toolkit/ArcGISToolkit/JobManager.swift
nixta: Yeah. I had wondered about this, but since it has worked until now, I didn't want to change it. My assumption was that the enumerator is unaffected by the removals.

However, even if that's the case, this is also better since it will invoke writing to userDefaults less often (once for the entire collection of jobs, rather than once per change). I think.

Toolkit/ArcGISToolkit/JobManager.swift
nixta: Ooh. I was not aware of LazyCollections. Thank you!

Toolkit/ArcGISToolkit/JobManager.swift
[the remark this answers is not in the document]
philium: You discovered the workaround and the warning that prompted it. Definitely seems like a bug; `self` shouldn't be needed there. The same warning is generated when using Swift 5 as well.

nixta:
@philium Please re-review. Have implemented most of your changes. Because of the behavior that `didSet` is called after `super.init()` in the initializer, I still had to maintain the `suppressSaveToUserDefaults`, but it's been refactored out of the way.

Toolkit/ArcGISToolkit/JobManager.swift
```diff
@@ -55,80 +55,57 @@ public class JobManager: NSObject {
...
-    private func instantiateStateFromJSON(json: JSONDictionary){
-        for (jobID, value) in json{
-            if let jobJSON = value as? JSONDictionary{
-                if let job = (try? AGSJob.fromJSON(jobJSON)) as? AGSJob{
-                    self.keyedJobs[jobID] = job
-                }
-            }
-        }
+    deinit {
+        keyedJobs = [:]
```
philium: Is this intended to stop observing the the jobs? As far as I understand, property observers are not called when the property is set from inside `deinit`. However, I recall iOS 11 / macOS 10.13 introducing some changes to KVO where remaining observers after the object is deallocated will be cleaned up by KVO. That may be happening here.

That being said, I just had another idea. If you use the newer Swift closure-based KVO API, that could simplify things somewhat. [Here](https://github.com/Esri/arcgis-runtime-toolkit-ios/commit/e0f039ae083dccd1ba8bcb5b0f581787eed4160c) is what that would look like.

nixta: I just realized deinit wasn't getting called in my tests, so this does a very bad thing. I'll back this change out.

nixta: And yes, KVO is resilient to not having observers removed now, but still not a bad practice to explicitly do it.

Issue #64, opened 2019-07-24
Row for previously selected sample remains selected


philium:
When a selected example is dismissed, the row for that example remains selected. This is different than the default behavior of `UITableViewController`, which deselects the currently selected row on `viewWillAppear(_:)`.

![Persistent Selection](https://user-images.githubusercontent.com/2257493/61823408-a0ae5480-ae10-11e9-8b0b-81c9b411f44f.gif)

PR #65, opened 2019-07-24
Make VCListViewController subclass UITableViewController

philium:
Fixes issue https://github.com/Esri/arcgis-runtime-toolkit-ios/issues/64 by using the default selection behavior provided by `UITableViewController`.

UITableViewController Selection

### DIFF
--- a/Examples/ArcGISToolkitExamples/VCListViewController.swift
+++ b/Examples/ArcGISToolkitExamples/VCListViewController.swift
@@ -14,7 +14,7 @@
 import UIKit
 import ArcGISToolkit
 
-open class VCListViewController: TableViewController {
+open class VCListViewController: UITableViewController {
     
     public var storyboardName: String?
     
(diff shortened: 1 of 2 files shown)


Examples/ArcGISToolkitExamples/VCListViewController.swift
```diff
@@ -25,30 +25,17 @@ open class VCListViewController: TableViewController {
...
-        super.init(nibName: nil, bundle: nil)
-    }
-    
     override open func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
         return viewControllerInfos.count
     }
     
     override open func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
-        let cell = tableView.dequeueReusableCell(withIdentifier: cellReuseIdentifier)!
+        let cell = tableView.dequeueReusableCell(withIdentifier: "cell", for: indexPath)
```
nixta: Why not make use of `cellReuseIdentifier`?

philium: Since the class no longer inherits from `TableViewController`, `TableViewController. cellReuseIdentifier` is no longer available.

We could define our our property for the reuse identifier, but since it is only needed in one place (since the storyboard handles registering the cell), it doesn't make much difference either way.

nixta: Oh :/ I hadn't noticed it stopped inheriting. I'd better pay more attention.

PR #67, opened 2019-07-31
Mhd/ar example updates

mhdostal:
- Add ability to switch Scenes via a toolbar button and popover.
- Add a status VC that displays relevant information about the AR session and app.
- Visualize planes found by ARKit.

FunctionalTestApp


mhdostal:
@philium @sbaskaran There were a couple of things I wanted to do with this but couldn't figure out:

- add a blurry background to the ARStatusViewController (and maybe the "Change Scene" popover).  Not sure if that's possible or if I should just set the transparency on the background.
- make the Toolbar transparent;  tired setting `isTransparent` and a background color with an alpha value, but no luck.

If you have any ideas, let me know.  Thanks.

Examples/ArcGISToolkitExamples/Misc/OptionsTableViewController.swift
```diff
@@ -0,0 +1,86 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import UIKit
+
+/// A basic interface for selecting one options from a list,
+/// showing the checkmark accessory for the selected cell.
+class OptionsTableViewController: UITableViewController {
```
philium: FYI: There are some substantive issues with this view controller. For example, rows aren't highlighted when touched. I plan on fixing that and making other improvements to this view controller at some point.

Examples/ArcGISToolkitExamples/ARExample.swift
```diff
@@ -35,15 +84,62 @@ class ARExample: UIViewController {
...
+        let statusItem = UIBarButtonItem(title: "Status", style: .plain, target: self, action: #selector(showStatus(_:)))
+
+        toolbar.setItems([UIBarButtonItem(barButtonSystemItem: .flexibleSpace, target: nil, action: nil),
+                          sceneItem,
+                          UIBarButtonItem(barButtonSystemItem: .flexibleSpace, target: nil, action: nil),
+                          statusItem], animated: false)
+        
+        // Add the status view and setup constraints.
+        if let controller = statusViewController {
+            view.addSubview(controller.view)
```
philium: You're adding the view controller's view as a subview, but you're not making the view controller a child. That works, but is effectively no different than if `controller` were just a view, not a view controller. I would suggest either making this a [container view controller](https://developer.apple.com/library/archive/featuredarticles/ViewControllerPGforiPhoneOS/ImplementingaContainerViewController.html#//apple_ref/doc/uid/TP40007457-CH11-SW12) by adding `controller` as a child view controller or making `ARStatusViewController` into `ARStatusView`.

Examples/ArcGISToolkitExamples/ARExample.swift
```diff
@@ -18,14 +18,63 @@ import ArcGIS
...
+    
+    /// The `ArcGISARView` that displays the camera feed and handles ARKit functionality.
+    private let arView = ArcGISARView(renderVideoFeed: true, tryUsingARKit: true)
+    
+    /// Denotes whether we've performed a hit test yet.
+    private var didHitTest: Bool = false
+
+    // View controller displaying current status of `ARExample`.
+    private let statusViewController: ARStatusViewController? = {
+        let storyBoard = UIStoryboard(name: "ARStatusViewController", bundle: nil)
```
philium: When a storyboard contains only one view controller, I recommend using a xib instead. It is much easier to instantiate (`ARStatusViewController()`, assuming the xib shares the same name as the view controller) and makes it clear that the view controller isn't part of a larger workflow. Yes, static table view cells are not supported in a xib, but I would recommend making the the table view's data source more robust anyway. Right now, if another scene were to be added, that would require a change to this view controller, `ARStatusViewController`, and the storyboard. Ideally only this view controller or `ARStatusViewController` would need to change, depending on who is responsible for populating the backing source used by the table view's data source.

Examples/ArcGISToolkitExamples/ARExample.swift
```diff
@@ -53,26 +149,87 @@ class ARExample: UIViewController {
...
-        let elevationSource = AGSArcGISTiledElevationSource(url: URL(string: "https://elevation3d.arcgis.com/arcgis/rest/services/WorldElevation3D/Terrain3D/ImageServer")!)
-        let surface = AGSSurface()
-        surface.elevationSources = [elevationSource]
-        surface.name = "baseSurface"
-        surface.isEnabled = true
-        surface.backgroundGrid.isVisible = false
-        scene.baseSurface = surface
+        // A view controller allowing the user to select the scene to show.
+        // Note: the `OptionsTableViewController` is copied from the "ArcGIS Runtime SDK for iOS Samples" code, found here:  https://github.com/Esri/arcgis-runtime-samples-ios
+        let controller = OptionsTableViewController(labels: sceneLabels, selectedIndex: selectedIndex) { [weak self] (newIndex) in
```
philium: `OptionsTableViewController` is handy in a navigation stack. But when presenting a list modally, I recommend using an action sheet instead. The currently selected action can be indicated by setting `isEnabled` to `false`.

Examples/ArcGISToolkitExamples/ARExample.swift
```diff
@@ -18,14 +18,63 @@ import ArcGIS
...
         arView.arSCNViewDelegate = self
         
+        // Set ourself as touch delegate so we can get touch events.
+        arView.sceneView.touchDelegate = self
+        
+        // Add our graphics overlay to the sceneView.
+        arView.sceneView.graphicsOverlays.add(graphicsOverlay)
+        
+        // Observe the `cameraController.translationFactor` property and update status when it changes.
+        translationFactorObservation = arView.observe(\ArcGISARView.translationFactor, options: [.initial, .new]){ [weak self] arView, change in
```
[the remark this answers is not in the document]
mhdostal: I want to be notified of the initial value so I can set the translationFactor property in the StatusViewController.  Without using `.initial`, that doesn't happen.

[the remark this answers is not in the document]
philium: Yes. That's what passing the `initial` option gives you, which I included in my suggested change. The `new` option causes the `newValue` property of `change` to be populated. Unless you're reading that new value, passing the `new` option just caused extra work to be done for no reason. Does that make sense?

Examples/ArcGISToolkitExamples/Misc/OptionsTableViewController.swift
mhdostal: Using UIAlertViewController as an action sheet.

Examples/ArcGISToolkitExamples/ARExample.swift
mhdostal: I made the change to a .xib.

Examples/ArcGISToolkitExamples/ARExample.swift
mhdostal: I am now calling `addChild(statusViewController)` from my parent view controller.  I'm still unclear on the reason or benefit.

mhdostal:
@philium I made all the requested changes; I got the blurry background to work on the tableView, thanks.

Regarding the Toolbar, I was hoping to get the transparency the same as the attribute view, as it looks a little funny having a more transparent bar underneath the toolbar.  I did NOT get this to work yet.

Please re-review.

mhdostal:
Updated UI

Examples/ArcGISToolkitExamples/Misc/ARStatusTableViewController.swift
```diff
@@ -0,0 +1,146 @@
...
+        return 1
+    }
+    
+    override func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+        return (section == 0) ? cellLabels.count : 0
+    }
+    
+    override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        // Don't reuse cells, as our table is essentially static.
+        let statusCell = UITableViewCell(style: .value1, reuseIdentifier: nil)
```
philium: Creating a new cell each time circumvents the cell queueing mechanism built into `UITableView`. And since the table view is being reloaded each time a property changes, that means that five old cells are discarded and five new cells are created rather than the same five cells being reused each time. I would suggest either creating a xib for the cell or a `UITableViewCell` subclass and registering that with a reuse identifier with the table view in `viewDidLoad()`. Then, dequeue that cell here using `dequeueReusableCell(withIdentifier:for:)`.

Side Note: You could register `UITableViewCell` with a reuse identifier, but since the cell style is passed as part of the initializer and it can't be changed, that would always get you the basic style. So either a xib or a subclass (requesting the desired style when calling `super`) is needed to get a non-basic cell.

Examples/ArcGISToolkitExamples/Misc/ARStatusTableViewController.swift
```diff
@@ -0,0 +1,146 @@
...
+        statusCell.backgroundColor = .clear
+        statusCell.textLabel?.font = UIFont.systemFont(ofSize: 12.0)
+        statusCell.detailTextLabel?.font = UIFont.boldSystemFont(ofSize: 12.0)
+        statusCell.detailTextLabel?.textColor = .black
+        
+        statusCell.textLabel?.text = cellLabels[indexPath.row]
+        
+        var detailString = ""
```
philium: Since this property is being initialized in every case in the following `switch` statement, you don't have to initialize it here and you could make it immutable:

```swift
let detailString: String
```

Examples/ArcGISToolkitExamples/ARExample.swift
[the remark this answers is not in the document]
mhdostal: Will just passing `.initial` still notify me of changes to the value?

mhdostal:
@philium I reverted to the storyboard version of the ARStatusViewController and made the other suggested changes.  Please re-review.

The one remaining issue I had was if it's possible to get the Toolbar opacity to match that of the Attribution bar, or do you think that's not something I should do.  Thanks.

Examples/ArcGISToolkitExamples/ARExample.swift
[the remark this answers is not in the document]
philium: Yes. You'll be notified of changes to the value no matter what, but specifying `initial` also means that you'll be notified right away:

> If specified, a notification should be sent to the observer immediately, before the observer registration method even returns.

If passing `initial` meant that you only got the initial value, it would be much simpler to just read the value directly.

Examples/ArcGISToolkitExamples/Misc/ARStatusViewController.swift
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        // Add a blur effect behind the table view.
+        tableView.backgroundView = UIVisualEffectView(effect: UIBlurEffect(style: .regular))
+    }
+
+    // MARK: - Table view data source
```
philium: These data source methods aren't necessary. One of the advantages of using static cells is that the storyboard is the data source, so the system already knows how many sections, how many rows, and which cells to dequeue for each row.

Examples/ArcGISToolkitExamples/Misc/ARStatusViewController.storyboard
```diff
@@ -0,0 +1,199 @@
...
+                <tableViewController id="4p9-jS-sat" customClass="ARStatusViewController" customModule="ArcGISToolkitExamples" customModuleProvider="target" sceneMemberID="viewController">
+                    <tableView key="view" clipsSubviews="YES" contentMode="scaleToFill" alwaysBounceVertical="YES" dataMode="static" style="plain" separatorStyle="none" allowsSelection="NO" rowHeight="22" estimatedRowHeight="-1" sectionHeaderHeight="28" sectionFooterHeight="28" id="bjp-Dc-dPX">
+                        <rect key="frame" x="0.0" y="0.0" width="414" height="896"/>
+                        <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
+                        <color key="backgroundColor" white="0.0" alpha="0.0" colorSpace="custom" customColorSpace="genericGamma22GrayColorSpace"/>
+                        <sections>
+                            <tableViewSection id="Nc7-BF-joG">
+                                <cells>
+                                    <tableViewCell clipsSubviews="YES" contentMode="scaleToFill" preservesSuperviewLayoutMargins="YES" selectionStyle="default" indentationWidth="10" id="MqH-RN-jot" userLabel="Tracking State">
```
philium: The cells look fine in portrait, but rotating to landscape reveals a problem:

image

An easy solution would be to just use Right Detail style cells.  Here's the table with the Translation Factor row updated to use a non-custom cell:

image

philium:
About the toolbar background. I played around with it for a little while and couldn't find a way to do what you want. That doesn't mean it isn't possible, just that I don't know how. Sorry.

Examples/ArcGISToolkitExamples/Misc/ARStatusViewController.storyboard
mhdostal: I have never seen that; it works OK with all orientations on my two devices.  The `accessoryView` detail labels are set up to be right-justified;  they do not have any constraints associated with them because of how they are setup.

What device/OS version are you running on?

mhdostal:
@philium Made all the changes; I haven't seen the issues with the StatusVC and device rotation, so no sure what's up with that.

Examples/ArcGISToolkitExamples/Misc/ARStatusViewController.storyboard
philium: I'm using an iPhone XR simulator running iOS 12.4. The issue is that the rows are being inset from the safe area on the right. The detail labels on the right side of the cell are sized from the storyboard and, due to their width, they are pushing the labels on the left further to the left.

mhdostal: Ok, got it.  Made the changes, including changing the cells to Right Detail cells and adjusted the constraints to the safe area.

PR #69, opened 2019-09-04
Mhd/ar enhancements

mhdostal:
Enhancements and bug fixes, including, but not limited to:

- When continuous GPS is enabled it would be a nice if we hid the elevation calibration slider
- Calibration should be disabled in Table top
- Suppress user directions view message: "Keep moving your device" when using continuous GPS locations. This is because with every location update, ARKit is reset, so it's perpetually in a state of initialization.
- Disable changing scenes if we're calibrating.
- Pass LocationDataSourceDelegate methods through to client
- add horizontalAccuracy, verticalAccuracy and LDS status to StatusVC

### DIFF
--- a/Examples/ArcGISToolkitExamples/ARExample.swift
+++ b/Examples/ArcGISToolkitExamples/ARExample.swift
@@ -67,6 +67,12 @@ class ARExample: UIViewController {
     /// The toolbar used to display controls for calibration, changing scenes, and status.
     private var toolbar = UIToolbar(frame: .zero)
     
+    /// Button used to display the `CalibrationView`.
+    private let calibrationItem = UIBarButtonItem(title: "Calibration", style: .plain, target: self, action: #selector(displayCalibration(_:)))
+    
+    /// Button used to change the current scene.
+    private let sceneItem = UIBarButtonItem(title: "Change Scene", style: .plain, target: self, action: #selector(changeScene(_:)))
+
     // MARK: Initialization
     
     override func viewDidLoad() {
@@ -181,6 +187,9 @@ class ARExample: UIViewController {
         
         // Hide directions view if we're calibrating.
         userDirectionsView.isHidden = startCalibrating
+        
+        // Disable changing scenes if we're calibrating.
+        sceneItem.isEnabled = !startCalibrating
     }
     
     /// Allow users to change the current scene.
@@ -209,6 +218,12 @@ class ARExample: UIViewController {
                 self.arView.startTracking(useLocationDataSourceOnce: info.useLocationDataSourceOnce, completion: { [weak self] (error) in
                     self?.statusViewController?.errorMessage = error?.localizedDescription
                 })
+                
+                // Disable elevation control if we're using continuous GPS.
+                self.calibrationView?.elevationControlVisibility = info.useLocationDataSourceOnce
+                
+                // Disable calibration if we're in table top
+                self.calibrationItem.isEnabled = !info.tableTop
                 
                 // Reset didPlaceScene variable
                 self.didPlaceScene = false
@@ -244,12 +259,6 @@ class ARExample: UIViewController {
             toolbar.trailingAnchor.constraint(equalTo: view.trailingAnchor),
             toolbar.bottomAnchor.constraint(equalTo: arView.sceneView.attributionTopAnchor)
             ])
-        
-        // Create a toolbar button for calibration.
-        let calibrationItem = UIBarButtonItem(title: "Calibration", style: .plain, target: self, action: #selector(displayCalibration(_:)))
-        
-        // Create a toolbar button to change the current scene.
-        let sceneItem = UIBarButtonItem(title: "Change Scene", style: .plain, target: self, action: #selector(changeScene(_:)))
         
         // Create a toolbar button to display the status.
         let statusItem = UIBarButtonItem(title: "Status", style: .plain, target: self, action: #selector(showStatus(_:)))
--- a/Examples/ArcGISToolkitExamples/ARExample.swift
+++ b/Examples/ArcGISToolkitExamples/ARExample.swift
@@ -133,7 +133,9 @@ class ARExample: UIViewController {
                                       (sceneFunction: emptyScene, label: "Empty - Full Scale", tableTop: false, trackingMode: .initial)])
         
         // Use the first sceneInfo to create and set the scene.
-        selectSceneInfo(sceneInfo.first)
+        if let info = sceneInfo.first {
+            selectSceneInfo(info)
+        }
         
         // Debug options for showing world origin and point cloud scene analysis points.
 //        arView.arSCNView.debugOptions = [.showWorldOrigin, .showFeaturePoints]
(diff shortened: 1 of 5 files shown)


Examples/ArcGISToolkitExamples/Misc/ARStatusViewController.swift
```diff
@@ -13,7 +13,7 @@
 
 import UIKit
 import ARKit
-
```
philium: I would suggest keeping the blank line between the import statements and the extension.

Examples/ArcGISToolkitExamples/Misc/ARStatusViewController.swift
```diff
@@ -97,6 +115,39 @@ class ARStatusViewController: UITableViewController {
         }
     }
 
+    /// The horizontal accuracy of the last location.
+    public var horizontalAccuracy: Double = 1.0 {
```
philium: If you wanted to get fancy, you could make these measurements:

```swift
public var horizontalAccuracyMeasurement = Measurement(value: 1, unit: UnitLength.meters)
```

And then use a measurement formatter to convert the measurement to a string with the appropriate units. That's typically how we do this sort of thing in the samples app (at least for more recent samples).

Examples/ArcGISToolkitExamples/Misc/ARStatusViewController.swift
```diff
@@ -97,6 +115,39 @@ class ARStatusViewController: UITableViewController {
...
+        didSet {
+            guard horizontalAccuracyLabel != nil else { return }
```
philium: A couple things:
1. If this was set from a background queue, the label could become non-nil by the time the work would have been dispatched to the main queue.
2. You could just use optional chaining:

    ```swift
    self.horizontalAccuracyLabel?.text = String(format: "%.0f", self.horizontalAccuracy)
    ```

Same with the other new properties.

Examples/ArcGISToolkitExamples/Misc/CalibrationView.swift
```diff
@@ -14,15 +14,21 @@
...
+    /// Denotes whether to show the elevation control and label; defaults to `true`.
+    public var elevationControlVisibility: Bool = true {
+        didSet {
+            elevationSlider.isHidden = !elevationControlVisibility
+            elevationLabel.isHidden = !elevationControlVisibility
+        }
+    }
+    
+    /// The `ArcGISARView` containing the origin camera we will be updating.
+    private var arcgisARView: ArcGISARView!
```
philium: Correct me if I'm wrong, but this is only ever used to get at the origin camera, correct? If so, could you just initialize a calibration view with a camera instead? That way you don't need to introduce `ArcGISToolkit` as a dependency that is only used in passing.

Examples/ArcGISToolkitExamples/Misc/ARStatusViewController.swift
mhdostal: Inadvertent deletion...

Examples/ArcGISToolkitExamples/Misc/ARStatusViewController.swift
mhdostal: There's no reason this same pattern couldn't be applied to all the properties, so that's what I did.  Thanks!

Examples/ArcGISToolkitExamples/Misc/CalibrationView.swift
mhdostal: It also needs to set the `arcgisARView.originCamera`.

Examples/ArcGISToolkitExamples/Misc/CalibrationView.swift
```diff
@@ -52,23 +58,23 @@ class CalibrationView: UIView {
     }()
     
     /// The last elevation slider value.
-    var lastElevationValue: Float = 0
+    private var lastElevationValue: Float = 0
     
     // The last heading slider value.
-    var lastHeadingValue: Float = 0
+    private var lastHeadingValue: Float = 0
```
pgruenler: Does the comment above this need another slash?

mhdostal: This property and the next were obsolete and not used, so I removed them.

mhdostal:
@philium @pgruenler I've made the changes request and added the last bit of API design changes based on the final (hopefully) design.  Please take another look.

Examples/ArcGISToolkitExamples/ARExample.swift
```diff
@@ -175,8 +186,44 @@ class ARExample: UIViewController {
...
+        
+        // Disable changing scenes if we're calibrating.
+        sceneItem.isEnabled = !startCalibrating
     }
-
+    
+    /// Sets up the required functionality in order to display the scene and AR experience represented by `sceneInfo`.
+    ///
+    /// - Parameter sceneInfo: The sceneInfo used to set up the scene and AR experience.
+    fileprivate func selectSceneInfo(_ sceneInfo: SceneInfoType?) {
```
philium: This is not a big deal, but I would recommend making the parameter non-optional since the function does nothing when it is `nil`. Having an optional `SceneInfoType` should be the caller's problem, not the callee's problem.

Examples/ArcGISToolkitExamples/ARExample.swift
```diff
@@ -175,8 +186,44 @@ class ARExample: UIViewController {
...
+        guard let info = sceneInfo else { return }
+        
+        // Set currentSceneInfo to the selected scene info.
+        self.currentSceneInfo = info
```
philium: Now that this is no longer inside a closure, you could get rid of all the explicit references to `self`, if you want to.

Examples/ArcGISToolkitExamples/ARExample.swift
```diff
@@ -664,3 +692,17 @@ extension AGSScene {
     }
 }
 
+// MARK: AGSLocationChangeHandlerDelegate methods
+extension ARExample: AGSLocationChangeHandlerDelegate {
+    public func locationDataSource(_ locationDataSource: AGSLocationDataSource, locationDidChange location: AGSLocation) {
+        // When we get a new location, update the status view controller with the new horizontal and vertical accuracy.
+        statusViewController?.horizontalAccuracyMeasurement = Measurement(value: location.horizontalAccuracy, unit: UnitLength.meters)
```
philium: If you want to, you could just assign to the `value` property directly:

```swift
statusViewController?.horizontalAccuracyMeasurement.value = location.horizontalAccuracy
```

Toolkit/ArcGISToolkit/AR/ArcGISARView.swift
```diff
@@ -485,21 +497,23 @@ extension ArcGISARView: AGSLocationChangeHandlerDelegate {
...
             }
         }
         
-        // Always set originCamera; then reset ARKit
-        let oldCamera = cameraController.originCamera
-            
+        // Always set originCamera; then reset ARKit            
         // Create a new camera based on our location and set it on the cameraController.
-        if originCamera == nil {
+        // Note for the .initial tracking mode (or if we've yet to set an initial locatin),
```
pgruenler: typo: locatin -> location

PR #70, opened 2019-09-13
Suggested edits

nixta:
Just suggested a couple of edits for consistency, and I think the "**Bold**:" works nicely for the bullet list.

A couple of other thoughts:

1. You mention "It uses a user-provided AGSLocationDataSource for getting an initial GPS location and when continuous GPS tracking is required" but should we mention that this is the default? It sounds as if the user will need to create one of these and provide it (I assume that's not usually the case).
2. This seems like it could be restructured to be a little clearer: "You must also add the following entries to your application's Info.plist file to support use of the camera (for the live video feed) and, when using the AGSCLLocationDataSource, the GPS (for determining your device's location)".
3. For the `info.plist` keys, should we also list the actual literals (like `NSLocationWhenInUseUsageDescription`) and link to the [Apple doc](https://developer.apple.com/documentation/bundleresources/information_property_list/nslocationwheninuseusagedescription)?

### DIFF
--- a/Documentation/AR/README.md
+++ b/Documentation/AR/README.md
@@ -3,17 +3,17 @@
 Augmented reality experiences are designed to "augment" the physical world with virtual content that respects real world scale, position, and orientation of a device. In the case of Runtime, a SceneView contains 3D geographic data as virtual content on top of a camera feed which represents the real, physical world.
 
 The Augmented Reality (AR) toolkit component allows quick and easy integration of AR into your application for a wide variety of scenarios.  The toolkit recognizes the following common patterns for AR: 
-* Flyover – Flyover AR is a kind of AR scenario that allows you to explore a scene using your device as a window into the virtual world. A typical flyover AR scenario will start with the scene’s virtual camera positioned over an area of interest. You can walk around and reorient the device to focus on specific content in the scene. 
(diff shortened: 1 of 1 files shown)


mhdostal:
> 1. You mention "It uses a user-provided AGSLocationDataSource for getting an initial GPS location and when continuous GPS tracking is required" but should we mention that this is the default? It sounds as if the user will need to create one of these and provide it (I assume that's not usually the case).

The user *will* need to create and supply one.

mhdostal:
> * This seems like it could be restructured to be a little clearer: "You must also add the following entries to your application's Info.plist file to support use of the camera (for the live video feed) and, when using the AGSCLLocationDataSource, the GPS (for determining your device's location)".
> * For the `info.plist` keys, should we also list the actual literals (like `NSLocationWhenInUseUsageDescription`) and link to the [Apple doc](https://developer.apple.com/documentation/bundleresources/information_property_list/nslocationwheninuseusagedescription)?

@nixta Agreed.  I'll make those changes as well.

nixta:
> The user will need to create and supply one.

Oh. Ah. Right. There’s no `AGSLocationDisplay` in a scene view to steal it from. I see.

PR #71, opened 2019-09-13
Mhd/release stuff

mhdostal:
Adds an READ.ME file for the AR components and adds the name of the current Scene as the title of the ARExample ViewController.


Documentation/AR/README.md
```diff
@@ -0,0 +1,53 @@
+# AR
+
+Augmented reality experiences are designed to "augment" the physical world with virtual content that respects real world scale, position, and orientation of a device. In the case of Runtime, a SceneView contains 3D geographic data as virtual content on top of a camera feed which represents the real, physical world.
```
philium:
```suggestion
Augmented reality experiences are designed to "augment" the physical world with virtual content that respects real world scale, position, and orientation of a device. In the case of Runtime, a SceneView displays 3D geographic data as virtual content on top of a camera feed which represents the real, physical world.
```

mhdostal:
@philium Made a small change to how we hide the ARKit SCNView.  The `arSCNView` needs to stay in the view hierarchy in order for ARKit to work, so set it's alpha to 0.0 instead.

PR #74, opened 2019-09-25
Mhd/swift lint

mhdostal:
I've added SwiftLint to the Toolkit and ToolkitExamples project; I used the SwiftLint.yml file used in our internal test repo; it exists in both the Examples and Toolkit directories as I couldn't figure out how to reference it in another directory.

Most of the changes were made via `swiftlint autocorrect`, but there were a few that were manual and could be prone to error and should be reviewed more closely.  Specifically:

- change from `something.count > 0` to `!something.isEmpty`; check for inverted Bool result
- changes in KVO in the JobManager; this was probably the biggest change as now the KVO is block based and we need to hold on to the observation object
- some methods in the Scalebar and TimeSlider required disabling the `cyclomatic_complexity` check.  I chose to disable them instead of rewriting the methods.


Examples/.swiftlint.yml
```diff
@@ -0,0 +1,69 @@
+included:
```
philium: You could remove this line if you want since it isn't needed.

mhdostal:
FYI, I added SwiftLint information in the main Readme.

Examples/.swiftlint.yml
```diff
@@ -0,0 +1,69 @@
...
+  - function_body_length
+  - function_parameter_count
+  - identifier_name
+  - large_tuple
+  - line_length
+  - nesting
+  - todo
+  - trailing_whitespace
+  - type_body_length
+custom_rules:
```
philium: Unless you're planning to add a unit test target to this project, you could just get rid of this custom rule.

Examples/ArcGISToolkitExamples/VCListViewController.swift
```diff
@@ -15,12 +15,11 @@ import UIKit
 import ArcGISToolkit
 
 open class VCListViewController: UITableViewController {
-    
     public var storyboardName: String?
     
     public var viewControllerInfos: [(vcName: String, viewControllerType: UIViewController.Type, nibName: String?)] = [
-        ]{
-        didSet{
+        ] {
```
philium: What would you think of moving this up to the previous line? It looks a little odd here on its own.

README.md
```diff
@@ -50,6 +50,8 @@ Note that you must also have the __ArcGIS Runtime SDK for iOS__ installed and yo
  4. Drag the `ArcGISToolkit.framework` from the `ArcGISToolkit.xcodeproj/ArcGISToolkit/Products` folder to the "TARGETS" settings for your application and drop it in the "Embedded Binaries" section in the "General" tab
  5. Add `import ArcGISToolkit` in your source code and start using the toolkit components 
 
+## SwiftLint
+New in the 100.6.0 release is SwiftLint support for both the Toolkit and Examples app.  You will need to install SwiftLint from [here](https://github.com/realm/SwiftLint) in order to build.  The specific rules the linter uses can be found in the `swiftlint.yml` files in the `Toolkit` and `Examples` directories.
```
philium: > You will need to install SwiftLint… in order to build

The script produces a warning, not an error, when SwiftLint is missing, so users should still be able to build even if they don't have SwiftLint installed.

PR #80, opened 2019-11-14
Mhd/unit test integration

mhdostal:
Adds a UnitTest target to the ArcGISToolkit project.
Adds initial "component creation" test, simply verifying all components can be created

Provides the framework to add unit tests for all new and existing components.


mhdostal:
cc: @dg0yal

mhdostal:
> I suppose you may not have to change the path of the ArcGIS framework if you don’t want to. I didn’t try changing the path back to see if it still works.

I tried changing it back (to `"~/Library...`) but it failed, so I left it as-is.

PR #83, opened 2019-11-21
DataSource tests

mhdostal:
Small PR for the three new DataSource tests.  There is no test design yet, as it's still in progress.


mhdostal:
had more than just the tests; closing and will create a new PR.

PR #90, opened 2020-01-02
Fix Auto Layout issues with MeasureToolbar

philium:
The issues stemmed from the toolbar laying out its subviews before it had a valid size. The solution I used is to assign the toolbar's items in `layoutSubviews()`, which is only called after the toolbar is positioned in its super view. I also made the result view a bar item.

### DIFF
--- a/Toolkit/ArcGISToolkit/MeasureToolbar.swift
+++ b/Toolkit/ArcGISToolkit/MeasureToolbar.swift
@@ -37,6 +37,7 @@ class MeasureResultView: UIView {
                 valueLabel.text = helpText
                 unitButton.isHidden = true
                 unitButton.setTitle(nil, for: .normal)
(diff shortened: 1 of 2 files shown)


rolson:
I was just testing this. If I use the area mode:

image


and then tap the "acres" button, I still see linear units:

image

I don't see that in `master`

rolson:
to reproduce the bug:
- go to the measure feature mode:
image

- tap a line feature:

image

- go into measure area mode:

image

- tap the acres button

rolson:
Actually - just re-tested with master now that I know the exact steps and I see the issue in master as well. So I will approve this PR and create a separate issue for this bug.

philium:
@rolson Thank you for looking at this so carefully and finding that bug!

PR #93, opened 2020-01-03
Remove resume button from job manager sample

rolson:
should just do this in viewDidLoad.

### DIFF
--- a/Examples/ArcGISToolkitExamples/JobManagerExample.swift
+++ b/Examples/ArcGISToolkitExamples/JobManagerExample.swift
@@ -141,7 +141,14 @@ class JobManagerExample: TableViewController {
         tableView.register(JobTableViewCell.self, forCellReuseIdentifier: "JobCell")
         
         // resume any paused jobs
-        JobManager.shared.resumeAllPausedJobs(statusHandler: self.jobStatusHandler, completion: self.jobCompletionHandler)
+        JobManager.shared.resumeAllPausedJobs(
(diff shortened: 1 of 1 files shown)


Examples/ArcGISToolkitExamples/JobManagerExample.swift
```diff
@@ -139,6 +139,9 @@ class JobManagerExample: TableViewController {
         
         // job cell registration
         tableView.register(JobTableViewCell.self, forCellReuseIdentifier: "JobCell")
+        
+        // resume any paused jobs
+        JobManager.shared.resumeAllPausedJobs(statusHandler: self.jobStatusHandler, completion: self.jobCompletionHandler)
```
philium: I realize this isn't your code, but there's a subtle bug here. By passing methods as the closures, a strong reference to `self` is being retained, preventing the view controller from being deinitialized. I'm guessing that's why I got a prompt for credentials back on the main view:

image

Either way, the solution is to pass closures with `weak` references to `self` that call the appropriate methods rather than pass the methods themselves. The same is true wherever a job is started (lines `275` and `301`. If you don't want to fix it yourself, I can submit a PR after yours is merged.

rolson: good point - I can fix this - thank you

philium: Thank you for fixing that!

PR #94, opened 2020-01-04
use new background task framework for job manager

rolson:
As Mark pointed out, the old way of doing background fetch has been deprecated. This adds support for the new way for iOS 13 devices.

BTW - for iOS 12 devices the old way still works - you just need to use the Xcode menu: `Debug -> Simulate Background Fetch`.

For iOS 13 devices, there is no way to test this on the simulator. You have to use a device and the documentation for testing it is [here](https://developer.apple.com/documentation/backgroundtasks/starting_and_terminating_tasks_during_development?language=objc)

This builds on the PR from @philium [here](https://github.com/Esri/arcgis-runtime-toolkit-ios/pull/91). So that should be merged first.

### DIFF
--- a/Examples/ArcGISToolkitExamples.xcodeproj/project.pbxproj
+++ b/Examples/ArcGISToolkitExamples.xcodeproj/project.pbxproj
@@ -230,6 +230,7 @@
 				TargetAttributes = {
 					8839043A1DF6022A001F3188 = {
 						CreatedOnToolsVersion = 8.0;
+						DevelopmentTeam = P8HGHS7JQ8;
 						LastSwiftMigration = 1110;
 						ProvisioningStyle = Automatic;
 						SystemCapabilities = {
@@ -508,7 +509,7 @@
 				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES;
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
-				DEVELOPMENT_TEAM = "";
+				DEVELOPMENT_TEAM = P8HGHS7JQ8;
 				FRAMEWORK_SEARCH_PATHS = (
 					"$(inherited)",
 					"$(USER_LIBRARY_DIR)/SDKs/ArcGIS/iOS/Frameworks/Dynamic",
@@ -528,7 +529,7 @@
 				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES;
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
-				DEVELOPMENT_TEAM = "";
+				DEVELOPMENT_TEAM = P8HGHS7JQ8;
 				FRAMEWORK_SEARCH_PATHS = (
 					"$(inherited)",
 					"$(USER_LIBRARY_DIR)/SDKs/ArcGIS/iOS/Frameworks/Dynamic",
--- a/Examples/ArcGISToolkitExamples.xcodeproj/project.pbxproj
+++ b/Examples/ArcGISToolkitExamples.xcodeproj/project.pbxproj
@@ -230,7 +230,6 @@
 				TargetAttributes = {
 					8839043A1DF6022A001F3188 = {
 						CreatedOnToolsVersion = 8.0;
-						DevelopmentTeam = P8HGHS7JQ8;
 						LastSwiftMigration = 1110;
 						ProvisioningStyle = Automatic;
 						SystemCapabilities = {
@@ -509,7 +508,7 @@
 				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES;
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
-				DEVELOPMENT_TEAM = P8HGHS7JQ8;
+				DEVELOPMENT_TEAM = "";
 				FRAMEWORK_SEARCH_PATHS = (
 					"$(inherited)",
 					"$(USER_LIBRARY_DIR)/SDKs/ArcGIS/iOS/Frameworks/Dynamic",
@@ -529,7 +528,7 @@
 				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES;
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
-				DEVELOPMENT_TEAM = P8HGHS7JQ8;
+				DEVELOPMENT_TEAM = "";
 				FRAMEWORK_SEARCH_PATHS = (
 					"$(inherited)",
 					"$(USER_LIBRARY_DIR)/SDKs/ArcGIS/iOS/Frameworks/Dynamic",
--- a/Examples/ArcGISToolkitExamples/AppDelegate.swift
+++ b/Examples/ArcGISToolkitExamples/AppDelegate.swift
@@ -23,7 +23,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         
         if #available(iOS 13.0, *) {
             let result = JobManager.shared.registerForBackgroundUpdates()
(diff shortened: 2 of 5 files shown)


Examples/ArcGISToolkitExamples.xcodeproj/project.pbxproj
```diff
@@ -230,6 +230,7 @@
 				TargetAttributes = {
 					8839043A1DF6022A001F3188 = {
 						CreatedOnToolsVersion = 8.0;
+						DevelopmentTeam = P8HGHS7JQ8;
```
philium: Please remove the development team information.

Toolkit/ArcGISToolkit/JobManager.swift
```diff
@@ -198,6 +202,7 @@ public class JobManager: NSObject {
     /// - Parameters:
     ///   - application:  See [Apple's documentation](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623125-application)
     ///   - completionHandler:  See [Apple's documentation](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1623125-application)
+    @available(*, deprecated, message: "Please use `registerForBackgroundUpdates` instead")
```
philium:
```suggestion
    @available(iOS, deprecated: 13.0, message: "Please use 'registerForBackgroundUpdates()' instead")
```

Toolkit/ArcGISToolkit/JobManager.swift
```diff
@@ -212,6 +217,58 @@ public class JobManager: NSObject {
...
+    /// Registers a task for updating job status in the background.
+    /// You must add an entry in the app plist for com.esri.arcgis.toolkit.jobmanager.refresh under BGTaskSchedulerPermittedIdentifiers.
+    /// This must be called before the end of the app launch sequence.
+    /// This must be tested on device, does not work on the simulator. For debugging this, see documentation here: https://developer.apple.com/documentation/backgroundtasks/starting_and_terminating_tasks_during_development?language=objc
+    @available(iOS 13.0, *)
+    @discardableResult
+    public func registerForBackgroundUpdates() -> Bool {
+        // register the bg task
+        let result = BGTaskScheduler.shared.register(forTaskWithIdentifier: bgTaskIdentifier, using: nil) { task in
+            self.handleBackgroundRefresh(task: task as! BGAppRefreshTask)
```
philium: See my suggestion on line `246` which makes this suggestion possible.

```suggestion
            self.handleBackgroundRefresh(task: task)
```

Toolkit/ArcGISToolkit/JobManager.swift
```diff
@@ -212,6 +217,58 @@ public class JobManager: NSObject {
...
+            bgObserver = NotificationCenter.default.addObserver(forName: UIApplication.willResignActiveNotification, object: nil, queue: .main) { [weak self] _ in
+                self?.scheduleNextBackgroundRefresh()
+            }
+        }
+        
+        return result
+    }
+    
+    @available(iOS 13.0, *)
+    private func handleBackgroundRefresh(task: BGAppRefreshTask) {
```
philium: If you change the type of `task` to `BGTask`, you wouldn't need to cast on line `232`.

```suggestion
    private func handleBackgroundRefresh(task: BGTask) {
```

Toolkit/ArcGISToolkit/JobManager.swift
```diff
@@ -212,6 +217,58 @@ public class JobManager: NSObject {
...
+    
+    /// Registers a task for updating job status in the background.
+    /// You must add an entry in the app plist for com.esri.arcgis.toolkit.jobmanager.refresh under BGTaskSchedulerPermittedIdentifiers.
+    /// This must be called before the end of the app launch sequence.
+    /// This must be tested on device, does not work on the simulator. For debugging this, see documentation here: https://developer.apple.com/documentation/backgroundtasks/starting_and_terminating_tasks_during_development?language=objc
+    @available(iOS 13.0, *)
+    @discardableResult
+    public func registerForBackgroundUpdates() -> Bool {
+        // register the bg task
+        let result = BGTaskScheduler.shared.register(forTaskWithIdentifier: bgTaskIdentifier, using: nil) { task in
```
philium: This closure should weakly capture `self`.

Toolkit/ArcGISToolkit/JobManager.swift
```diff
@@ -212,6 +217,58 @@ public class JobManager: NSObject {
...
+    private let bgTaskIdentifier = "com.esri.arcgis.toolkit.jobmanager.refresh"
+    private var bgObserver: AnyObject?
+    
+    /// Registers a task for updating job status in the background.
+    /// You must add an entry in the app plist for com.esri.arcgis.toolkit.jobmanager.refresh under BGTaskSchedulerPermittedIdentifiers.
+    /// This must be called before the end of the app launch sequence.
+    /// This must be tested on device, does not work on the simulator. For debugging this, see documentation here: https://developer.apple.com/documentation/backgroundtasks/starting_and_terminating_tasks_during_development?language=objc
+    @available(iOS 13.0, *)
+    @discardableResult
+    public func registerForBackgroundUpdates() -> Bool {
```
philium: You might want to document the return value of this method.

Toolkit/ArcGISToolkit/JobManager.swift
rolson: The way you are recommending this is the way I actually wrote it as well!  The problem is I wasn't seeing the background refresh work and every sample I looked at had it written this way and I started wondering if there was something in apples code checking for the parameter type. I saw at least 3 examples of the code written like this - including apple's own sample code, so I changed it to this.

philium: Weird! I can't think of why that would make a difference, but apparently it does. I just saw that it compiled; I didn't actually test it. Thank you for the explanation.

rolson: I don't know if it did actually make a difference in the end. The way to test this is not great and I didn't realize it didn't work on the simulator at all until later.

Toolkit/ArcGISToolkit/JobManager.swift
rolson: yes - thank you

Examples/ArcGISToolkitExamples/JobManagerExample.swift
```diff
@@ -152,11 +139,42 @@ class JobManagerExample: TableViewController {
         
         // job cell registration
         tableView.register(JobTableViewCell.self, forCellReuseIdentifier: "JobCell")
+        
+        // resume any paused jobs
+        JobManager.shared.resumeAllPausedJobs(
```
philium: Now that there is a corresponding action when the view controller goes away, I'm wondering if a more appropriate place for resuming the jobs is `viewDidAppear(_:)` and the jobs should be paused in `viewWillDisappear(_:)`. That way any authentication challenges are only shown when the view controller is on screen.

rolson: that might be a better option. I can change to that.

Examples/ArcGISToolkitExamples/AppDelegate.swift
```diff
@@ -48,6 +53,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     // Here is where we forward background fetch to the JobManager
     // so that jobs can be updated in the background
     func application(_ application: UIApplication, performFetchWithCompletionHandler completionHandler: @escaping (UIBackgroundFetchResult) -> Void) {
+        // We only do it this way for pre-iOS 13. Otherwise we use BGTask above.
+        // This method doesn't get called for iOS 13 or later because we have an entry
+        // in the plist for BGTaskSchedulerPermittedIdentifiers.
         JobManager.shared.application(application: application, performFetchWithCompletionHandler: completionHandler)
```
philium: This generates a warning…sometimes. I had a screenshot of it, but I accidentally deleted it and now it isn't showing up. 😞

```suggestion
        if #available(iOS 13.0, *) {
            // Nothing to do here.
        } else {
            JobManager.shared.application(application: application, performFetchWithCompletionHandler: completionHandler)
        }
```

PR #95, opened 2020-01-07
job manager: use background task to keep app alive

rolson:
After speaking with Eric on the apps team. The way the apps team uses `UIApplication.shared.beginBackgroundTask` instead of the background fetch stuff. Background fetch can take hours to get called according to the Apple documentation. Instead `beginBackgroundTask` keeps the app alive for up to 10 minutes, giving most jobs the ability to finish or at least get to the out of process download phase.

### DIFF
--- a/Examples/ArcGISToolkitExamples/JobManagerExample.swift
+++ b/Examples/ArcGISToolkitExamples/JobManagerExample.swift
@@ -309,8 +309,8 @@ class JobManagerExample: TableViewController {
                     self?.jobStatusHandler(status: $0)
                 },
                 completion: { [weak self] in
-                    self?.endBackgroundTask(backgroundTaskIdentifier)
                     self?.jobCompletionHandler(result: $0, error: $1)
+                    self?.endBackgroundTask(backgroundTaskIdentifier)
                 }
             )
             
@@ -360,8 +360,8 @@ class JobManagerExample: TableViewController {
                         self?.jobStatusHandler(status: $0)
                     },
                     completion: { [weak self] in
-                        self?.endBackgroundTask(backgroundTaskIdentifier)
                         self?.jobCompletionHandler(result: $0, error: $1)
+                        self?.endBackgroundTask(backgroundTaskIdentifier)
                     }
                 )
                 
--- a/Examples/ArcGISToolkitExamples/JobManagerExample.swift
+++ b/Examples/ArcGISToolkitExamples/JobManagerExample.swift
@@ -175,6 +175,12 @@ class JobManagerExample: TableViewController {
         super.viewWillDisappear(animated)
     }
     
+    deinit {
+        // clear out background tasks that we started for the jobs
+        backgroundTaskIdentifiers.forEach { UIApplication.shared.endBackgroundTask($0) }
+        backgroundTaskIdentifiers.removeAll()
+    }
+    
     override func viewDidLayoutSubviews() {
         super.viewDidLayoutSubviews()
         if let toolbar = toolbar, toolbar.items == nil {
@@ -253,7 +259,7 @@ class JobManagerExample: TableViewController {
             // make sure we are still around...
             guard let self = self, let strongTask = task else {
                 // don't need to end the background task here as that
-                // would have been done in dealloc
+                // would have been done in deinit
                 return
             }
             
@@ -341,7 +347,7 @@ class JobManagerExample: TableViewController {
             // make sure we are still around...
             guard let self = self else {
                 // don't need to end the background task here as that
-                // would have been done in dealloc
+                // would have been done in deinit
                 return
             }
             
--- a/Examples/ArcGISToolkitExamples/JobManagerExample.swift
+++ b/Examples/ArcGISToolkitExamples/JobManagerExample.swift
@@ -241,6 +241,9 @@ class JobManagerExample: TableViewController {
     }
     
     func generateGDB(URL: URL, syncModel: AGSSyncModel, extent: AGSEnvelope?) {
+        // try to keep the app running in the background for this job if possible
+        let backgroundTaskIdentifier = self.startBackgroundTask()
+        
         let task = AGSGeodatabaseSyncTask(url: URL)
         
         // hold on to task so that it stays retained while it's loading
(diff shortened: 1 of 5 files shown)


Examples/ArcGISToolkitExamples/JobManagerExample.swift
```diff
@@ -293,13 +299,17 @@ class JobManagerExample: TableViewController {
             
             // register the job with our JobManager shared instance
             JobManager.shared.register(job: job)
+
+            // try to keep the app running in the background for this job if possible
+            let backgroundTaskIdentifier = self.startBackgroundTask()
```
eito: Consider starting the background task as soon as this method is triggered rather than waiting for `load` to finish. Since `load` is network dependent, it's possible the app goes to the background _before_ the load is done. If it does then this background task will never get started. The user likely doesn't know (nor should they) if the task is still loading or not.

rolson:
> When running on the device, starting a couple of tasks, then backgrounding the apps I was able to get the banner notification when a task completed, but no badges showed up on the app icon. Is that expected?

Good question. The `UNMutableNotificationContent` should just need to be configured to show a badge number if the user wanted that.

> The Toolkit code is simplified with this approach, but the Example code is a bit more complex and seems like it's going to be common to all apps which want to perform Job tasks in the background. Is it within the scope of the JobManager component to handle all the UIApplication.shared.endBackgroundTask and UIApplication.shared.beginBackgroundTask stuff, along with the backgroundTaskIdentifier management? All of that code seems to happen either just after a call to JobManager.shared.xxx or in response to a callback/handler from the JobManager.

It would be hard to do the exact same way because the job manager doesn't listen for the job completion and it actually doesn't start the jobs. That could be added through KVO - but I do believe that it could become somewhat difficult to manage. Although I spend a bit of time to try to do an implementation like that and see how it turns out.

Examples/ArcGISToolkitExamples/JobManagerExample.swift
```diff
@@ -309,6 +319,20 @@ class JobManagerExample: TableViewController {
         }
     }
     
+    func startBackgroundTask() -> UIBackgroundTaskIdentifier {
+        var backgroundTaskIdentifier: UIBackgroundTaskIdentifier = .invalid
+        backgroundTaskIdentifier = UIApplication.shared.beginBackgroundTask {
```
eito: Can you not just use `let backgroundTaskIdentifier = <call to begin>`? If background isn't supported the `beginBackgroundTask` call will return `.invalid`

rolson:
https://devtopia.esri.com/runtime/cocoa/issues/9305

Examples/ArcGISToolkitExamples/JobManagerExample.swift
```diff
@@ -326,13 +350,17 @@ class JobManagerExample: TableViewController {
                 
                 // register the job with our JobManager shared instance
                 JobManager.shared.register(job: job)
+
+                // try to keep the app running in the background for this job if possible
+                let backgroundTaskIdentifier = self.startBackgroundTask()
```
eito: same as above, consider adding this at the beginning of `takeOffline`. Apple's doc suggests calling this "as early as possible before starting your task"

> Call this method as early as possible before starting your task, and preferably before your app actually enters the background. The method requests the task assertion for your app asynchronously. If you call this method shortly before your app is due to be suspended, there is a chance that the system might suspend your app before that task assertion is granted.

Examples/ArcGISToolkitExamples/JobManagerExample.swift
```diff
@@ -326,13 +350,17 @@ class JobManagerExample: TableViewController {
...
                 
                 // start the job
                 job.start(
                     statusHandler: { [weak self] in
                         self?.jobStatusHandler(status: $0)
                     },
                     completion: { [weak self] in
+                        self?.endBackgroundTask(backgroundTaskIdentifier)
                         self?.jobCompletionHandler(result: $0, error: $1)
```
eito: I'm not sure how soon after a call to `endBackgroundTask` the system will suspend your app if it's in the background, you may want to call this as _late_ as possible when things are completed.

Examples/ArcGISToolkitExamples/JobManagerExample.swift
rolson: I did consider that. I agree with the benefits. A few cons perhaps:

- I would now need to make sure to call `endBackgroundTask` anytime it's short-circuited.
- I could not move the code into the job manager, as @mhdostal suggested

Examples/ArcGISToolkitExamples/JobManagerExample.swift
rolson: I wish... I can't because it's used in the beginBackgroundTask block. Not sure how apple designed this api to be used. I found the suggestion above recommended somewhere

Examples/ArcGISToolkitExamples/JobManagerExample.swift
rolson: good call

Examples/ArcGISToolkitExamples/JobManagerExample.swift
```diff
@@ -166,9 +168,33 @@ class JobManagerExample: TableViewController {
...
+        backgroundTaskIdentifiers.forEach { UIApplication.shared.endBackgroundTask($0) }
+        backgroundTaskIdentifiers.removeAll()
+        
         super.viewWillDisappear(animated)
     }
     
+    deinit {
+        // clear out background tasks that we started for the jobs
+        backgroundTaskIdentifiers.forEach { UIApplication.shared.endBackgroundTask($0) }
+        backgroundTaskIdentifiers.removeAll()
```
philium: Isn't this unnecessary since the object is about to be deallocated?

rolson: yep. Removing

Examples/ArcGISToolkitExamples/JobManagerExample.swift
rolson: fixed

Examples/ArcGISToolkitExamples/JobManagerExample.swift
rolson: fixed

rolson:
thank you @philium @mhdostal and @eito !

PR #101, opened 2020-03-12
Update ArcGISARView.swift

armandzwan:
Make `locationDataSource` conform to `AGSLocationDataSource` to make it possible to inject a mock `locationDataSource` for testing.

It is briefly mentioned here: https://github.com/Esri/arcgis-runtime-toolkit-ios/pull/57

> might be replaced by an AGSLocationDataSource; there is a pending design issue to make the necessary changes to that class in order to support what's necessary for AR (and other use cases).

I think it is time to do so, no need to restrict it to `AGSCLLocationDataSource`, because the fix on line 496 will still work if a `AGSCLLocationDataSource` is used as a `locationDataSource`:
`if let clLocationDataSource = locationDataSource as? AGSCLLocationDataSource {`


mhdostal:
@armandzwan - thank you for the fix.  This was an oversight on my part, it always should have been `AGSLocationDataSource`.

PR #113, opened 2020-10-15
Remove duplicate dependency

philium:
I noticed that the example app had the toolkit framework listed as a dependency twice:

Screen Shot 2020-10-15 at 10 07 41 AM

This PR removes the duplicate dependency.


rolson (approved):

mhdostal (approved):
Thanks for catching this @philium!

PR #116, opened 2020-11-06
Update pod name in README

### LINKED ISSUE #114: Update pod name in README
When the coco pod name is referenced in the README it says the pod name is:  `pod ArcGIS-Runtime-Toolkit-iOS` but it should be corrected to `pod 'ArcGIS-Runtime-Toolkit-iOS' `
The corrected pod name is the one you can directly copy and paste into your podfile, the current pod name in the README if you copy and paste would give you an error.

zkline101:
Closes #114

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ The ArcGIS Runtime SDK for iOS Toolkit contains components that will simplify yo
 
 To use Toolkit in your project:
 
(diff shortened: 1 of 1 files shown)


mhdostal (approved):
Nice catch @zkline101 !

PR #118, opened 2020-11-09
Fix add record crash

### LINKED ISSUE #117: Toolkit iOS: PopupController adding new record triggers app termination
Details of this issue are [here](https://devtopia.esri.com/runtime/cocoa/issues/9378).

zkline101:
Closes #117

### DIFF
--- a/Toolkit/ArcGISToolkit/TemplatePickerViewController.swift
+++ b/Toolkit/ArcGISToolkit/TemplatePickerViewController.swift
@@ -184,7 +184,8 @@ public class TemplatePickerViewController: TableViewController {
                     let indexPathRow = indexPath.row
                     
                     // The size of the table view is still not finalized at this point, so we check to make sure the section number of `indexPath` does exist in the `tableView`.
-                    if indexPathSection < self.tableView.numberOfSections, indexPathRow < self.tableView.numberOfRows(inSection: indexPathSection) {
+                    if indexPathSection < self.tableView.numberOfSections,
+                       indexPathRow < self.tableView.numberOfRows(inSection: indexPathSection) {
                         self.tableView.reloadRows(at: [indexPath], with: .automatic)
                     }
                 }
--- a/Toolkit/ArcGISToolkit/TemplatePickerViewController.swift
+++ b/Toolkit/ArcGISToolkit/TemplatePickerViewController.swift
@@ -179,6 +179,7 @@ public class TemplatePickerViewController: TableViewController {
                     // update info with swatch
                     infos[index].swatch = image
                     
+                    // The size of the table view is still not finalized at this point, so we can't use `reloadRows(...)`.
                     self.tableView.reloadData()
                 }
             }
--- a/Toolkit/ArcGISToolkit/TemplatePickerViewController.swift
+++ b/Toolkit/ArcGISToolkit/TemplatePickerViewController.swift
@@ -180,12 +180,10 @@ public class TemplatePickerViewController: TableViewController {
                     infos[index].swatch = image
                     
                     let indexPath = self.indexPath(for: info)
-                    let indexPathSection = indexPath.section
-                    let indexPathRow = indexPath.row
                     
                     // Make sure the new indexPath is valid before reloading the row.
-                    if indexPathSection < self.tableView.numberOfSections,
-                       indexPathRow < self.tableView.numberOfRows(inSection: indexPathSection) {
+                    if indexPath.section < self.tableView.numberOfSections,
+                       indexPath.row < self.tableView.numberOfRows(inSection: indexPath.section) {
                         self.tableView.reloadRows(at: [indexPath], with: .automatic)
                     }
                 }
--- a/Toolkit/ArcGISToolkit/TemplatePickerViewController.swift
+++ b/Toolkit/ArcGISToolkit/TemplatePickerViewController.swift
@@ -179,8 +179,12 @@ public class TemplatePickerViewController: TableViewController {
                     // update info with swatch
                     infos[index].swatch = image
                     
-                    // The size of the table view is still not finalized at this point, so we can't use `reloadRows(...)`.
-                    self.tableView.reloadData()
+                    let indexPath = self.indexPath(for: info)
+                    
+                    // The size of the table view is still not finalized at this point, so we check to make sure the section number of `indexPath` does exist in the `tableView`.
+                    if indexPath.section < self.tableView.numberOfSections {
+                        self.tableView.reloadRows(at: [indexPath], with: .automatic)
+                    }
                 }
             }
         }
--- a/Toolkit/ArcGISToolkit/TemplatePickerViewController.swift
+++ b/Toolkit/ArcGISToolkit/TemplatePickerViewController.swift
@@ -182,7 +182,7 @@ public class TemplatePickerViewController: TableViewController {
                     let indexPath = self.indexPath(for: info)
                     
                     // The size of the table view is still not finalized at this point, so we check to make sure the section number of `indexPath` does exist in the `tableView`.
-                    if indexPath.section < self.tableView.numberOfSections {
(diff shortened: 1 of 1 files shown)


Toolkit/ArcGISToolkit/TemplatePickerViewController.swift
```diff
@@ -179,9 +179,7 @@ public class TemplatePickerViewController: TableViewController {
                     // update info with swatch
                     infos[index].swatch = image
                     
-                    // reload index where that info currently is
-                    let indexPath = self.indexPath(for: info)
-                    self.tableView.reloadRows(at: [indexPath], with: .automatic)
+                    self.tableView.reloadData()
```
mhdostal, line 182:
```suggestion
                    // The size of the table view is still not finalized at this point, so we can't use `reloadRows(...)`.
                    self.tableView.reloadData()
```

mhdostal (changes requested):
Just added a comment to the `reloadData()` method explaining why the more efficient `reloadRows` method can't be used.

Toolkit/ArcGISToolkit/TemplatePickerViewController.swift
zkline101: Gotcha

mhdostal (approved):

Toolkit/ArcGISToolkit/TemplatePickerViewController.swift
```diff
@@ -179,9 +179,8 @@ public class TemplatePickerViewController: TableViewController {
                     // update info with swatch
                     infos[index].swatch = image
                     
-                    // reload index where that info currently is
-                    let indexPath = self.indexPath(for: info)
-                    self.tableView.reloadRows(at: [indexPath], with: .automatic)
+                    // The size of the table view is still not finalized at this point, so we can't use `reloadRows(...)`.
+                    self.tableView.reloadData()
```
rolson, line 183: `reloadRows` provides a better UX when just updating the swatch. Is it possible to conditionalize this fix? I don't remember the specifics of this particular function, but if it reloads the entire table view every time a swatch comes in, that seems to be inefficient.

Is the indexPath nil or invalid in the crash case? If so, can you just bail?

zkline101: Hmm when I was testing I don't think `indexPath` was ever nil, I think it was referring to an `indexPath` in the table that did not exist at the time. Let me double check that

mhdostal: The index path in question is calculated based on the feature table and feature template info and their locations in the respective data sources, so there should always be a valid indexPath.

The `loadInfosAndCreateSwatches` where the code in question resides is only called once from `viewDidLoad`, so the overhead of using `reloadData()` won't be too bad.

zkline101: So the method returning the `indexPath` returns a non-optional. But I checked and the `indexPath` is referring to a section one that is not available when it runs `reloadRows`

zkline101: If you try to find the number of rows in the section (using `tableView.numberOfRows`) that is assigned to `indexPath` it crashes because that section is out of bounds of the current table view.

zkline101: Alright let me check that out

rolson (approved):

Toolkit/ArcGISToolkit/TemplatePickerViewController.swift
```diff
@@ -179,8 +179,12 @@ public class TemplatePickerViewController: TableViewController {
                     // update info with swatch
                     infos[index].swatch = image
                     
-                    // The size of the table view is still not finalized at this point, so we can't use `reloadRows(...)`.
-                    self.tableView.reloadData()
+                    let indexPath = self.indexPath(for: info)
+                    
+                    // The size of the table view is still not finalized at this point, so we check to make sure the section number of `indexPath` does exist in the `tableView`.
+                    if indexPath.section < self.tableView.numberOfSections {
```
mhdostal, line 185: Won't we also want to check to make sure the number of rows < those in the section?

zkline101: Yep and do you mean this `indexPath.row < self.tableView.numberOfRows(inSection: indexPath.section)`

mhdostal: Yup.  It could be that the section was created, but no rows were populated yet.  This all seems like a timing issue to me.

zkline101: Yep it seems that way.

rolson: Everything should be happening on the main thread, including the response from createSwatch. So I'm thinking it's just that some swatches in this repro case came back immediately before the next iteration of the run loop when table view has a chance to ask the datasource how many sections. So, if the section is there I'm thinking the row would be as well. But it probably doesn't hurt to check.

Toolkit/ArcGISToolkit/TemplatePickerViewController.swift
```diff
@@ -182,7 +182,7 @@ public class TemplatePickerViewController: TableViewController {
                     let indexPath = self.indexPath(for: info)
                     
                     // The size of the table view is still not finalized at this point, so we check to make sure the section number of `indexPath` does exist in the `tableView`.
```
mhdostal, line 186:
```suggestion
                    // Make sure the new indexPath is valid before reloading the row.
```

Toolkit/ArcGISToolkit/TemplatePickerViewController.swift
```diff
@@ -182,7 +182,7 @@ public class TemplatePickerViewController: TableViewController {
...
-                    if indexPath.section < self.tableView.numberOfSections {
+                    if indexPath.section < self.tableView.numberOfSections, indexPath.row < self.tableView.numberOfRows(inSection: indexPath.section) {
```
mhdostal: The line got kind of long, so putting it on two lines is preferred.
```suggestion
                    if indexPath.section < self.tableView.numberOfSections,
                        indexPath.row < self.tableView.numberOfRows(inSection: indexPath.section) {
```

mhdostal (changes requested):

Toolkit/ArcGISToolkit/TemplatePickerViewController.swift
```diff
@@ -179,9 +179,15 @@ public class TemplatePickerViewController: TableViewController {
                     // update info with swatch
                     infos[index].swatch = image
                     
-                    // reload index where that info currently is
                     let indexPath = self.indexPath(for: info)
-                    self.tableView.reloadRows(at: [indexPath], with: .automatic)
+                    let indexPathSection = indexPath.section
+                    let indexPathRow = indexPath.row
```
rolson, line 184: I don't really see a reason for these variables to be declared. Just use `indexPath.section` and `indexPath.row` inline

Toolkit/ArcGISToolkit/TemplatePickerViewController.swift
rolson: After debugging this with Zach and looking at the code a bit more, I do think just checking to make sure that the index path exists is a valid fix for this issue.

rolson (approved):

Toolkit/ArcGISToolkit/TemplatePickerViewController.swift
zkline101: I added the section one because `indexPath.section` was being referred to twice and then I thought if I had a section one there should be a row one. But yeah your right don't really need them.

mhdostal (approved):

PR #139, opened 2021-03-05
Convert Toolkit to Swift Package

zkline101:
Note: this will be apart of Update 11.
Things done in this PR:
- Toolkit Project -> Toolkit SP
- Toolkit examples project now uses toolkit SP as a dependency


zkline101:
@mhdostal When you get a chance can you review this, a fair amount of changes but everything should work as normal. I am gonna make a separate PR for the README changes.

mhdostal (approved):

Examples/ArcGISToolkitExamples.xcodeproj/project.pbxproj
```diff
@@ -547,6 +473,17 @@
...
+
+/* Begin XCSwiftPackageProductDependency section */
+		79CA564325F1952D00E9FE3D /* ArcGISToolkit */ = {
+			isa = XCSwiftPackageProductDependency;
+			productName = ArcGISToolkit;
+		};
+		79CA564525F1953500E9FE3D /* ArcGISToolkit */ = {
+			isa = XCSwiftPackageProductDependency;
+			productName = ArcGISToolkit;
+		};
```
philium, line 485: This duplicate identical dependency is caused by the library being explicitly added to the Dependencies build phase:

image

Xcode seems to implicitly track Swift package dependencies, so the explicit dependency could be removed, eliminating this duplicate entry here.

zkline101: Yep let me fix that

philium (changes requested):
The [reference](https://github.com/Esri/arcgis-runtime-toolkit-ios/blob/0722b9110ed5296190137abab2967b9c217dcb82/Examples/ArcGISToolkitExamples.xcodeproj/project.pbxproj#L59) to the XCFramework at `$HOME/Library/SDKs/ArcGIS/Frameworks/ArcGIS.xcframework` should be removed. In fact, the `Frameworks` group could be removed as well.

image

mhdostal (approved):

Sources/ArcGISToolkit/Compass.swift
```diff
@@ -54,7 +54,7 @@ public class Compass: UIImageView {
         super.init(frame: .zero)
         
         // Set our image to the CompassIcon in the Assets
-        let bundle = Bundle(for: type(of: self))
+        let bundle = Bundle.module
         image = UIImage(named: "CompassIcon", in: bundle, compatibleWith: nil)
```
philium, line 58: Since the desired bundle is now just a static stored property on `Bundle` itself, it would be simpler just to refer to that:

```suggestion
        image = UIImage(named: "CompassIcon", in: .module, compatibleWith: nil)
```

philium (approved):

mhdostal (approved):

zkline101:
@mhdostal @philium Thank you for the reviews. I will merge this when I get the go-ahead and the last piece is to update the release tag after this is merged.

dg0yal:
@zkline101 We need to make a couple of changes to the swift package manifest at https://github.com/Esri/arcgis-runtime-toolkit-ios/blob/zach/toolkit-SPM/Package.swift#L39 . 
1) The final resting URL of the sdk swift package repo will be under the Esri org, it is in the ArcGIS org only temporarily as we set things up. So you need to change `https://github.com/ArcGIS/arcgis-runtime-ios` to `https://github.com/Esri/arcgis-runtime-ios`
2) Currently the dependency on the SDK is declared as an exact match on the version number. For cocoapods we took the approach of aligning the sdk only till the major.minor version, and letting the patch version be a wildcard so that we could get the latest patch of the sdk when users installed the toolkit. We should follow a similar approach for SPM (you'll have to research how)


Also, i have a question about this comment -https://github.com/Esri/arcgis-runtime-toolkit-ios/pull/139#pullrequestreview-606701382 . Does this mean the toolkit project no longer builds with the SDK that was installed manually as described in https://github.com/Esri/arcgis-runtime-toolkit-ios/tree/zach/toolkit-SPM#manual

zkline101:
@dg0yal Ok thanks gotcha. For your second point, we could use [upToNextMinor(from:)](https://developer.apple.com/documentation/swift_packages/package/dependency/requirement/2878234-uptonextminor). There is no more toolkit project, though there still is a toolkit examples project. The toolkit package no longer needs the SDK to be installed manually, it will automatically add it as a dependency. In this [PR](https://github.com/Esri/arcgis-runtime-toolkit-ios/pull/140) the instructions are changed to reflect that.

zkline101:
Items left to do before update 11 is released:
- `.upToNextMinor(from: "100.10.0")` -> `.upToNextMinor(from: "100.11.0")`
- increase the iOS minumum in the package manifest and the examples project

zkline101:
@mhdostal @philium if you would like to take another look this is ready. The only items left to do are listed above.

mhdostal:
@zkline101 - Make all the required changes, but then push them to the Update11 branch.  That way there's nothing left to do later.

zkline101:
@mhdostal Gotcha let me make final changes then merge this into `update11` branch

mhdostal:
> By the way, until the update 11 tag is created for the arcgis sdk ios repo this swift package will fail.

To get around that, for pre-U11 testing, you'd just have to change this line back to `100.10.0`, correct?

`.package(name: "ArcGIS", url: "https://github.com/Esri/arcgis-runtime-ios", .upToNextMinor(from: "100.11.0"))`

zkline101:
Yep and when you change that line you may want to close the package and re-open it so it loads with the new changes to the package manifest.

mhdostal (approved):

Package.swift
```diff
@@ -0,0 +1,54 @@
...
+    ],
+    dependencies: [
+        .package(name: "ArcGIS", url: "https://github.com/Esri/arcgis-runtime-ios", .upToNextMinor(from: "100.11.0"))
+    ],
+    targets: [
+        .target(
+            name: "ArcGISToolkit",
+            dependencies: ["ArcGIS"],
+            swiftSettings: [
+                .define("SPM")
```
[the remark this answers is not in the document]
zkline101: ah I see, did not know that let me make those edits

Package.swift
```diff
@@ -0,0 +1,54 @@
...
+
+ email: contracts@esri.com
+ */
+
+import PackageDescription
+
+let package = Package(
+    name: "ArcGISToolkit",
+    platforms: [
+        .iOS(.v13)
```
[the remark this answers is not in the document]
zkline101: I can address that in a separate PR

mhdostal (approved):

philium (approved):

zkline101:
Thanks @mhdostal and @philium !

PR #141, opened 2021-03-18
Edit scalebar README

zkline101:
For this piece of code in the scalebar README, the variable is declared as `scalebar`, not `sb`.


mhdostal (approved):
Nice catch!

PR #142, opened 2021-03-24
More Update 11 README updates

zkline101:
Found some more edits and @dg0yal pointed out that the Carthage banner is still present.


mhdostal (approved):

zkline101:
Thanks, @mhdostal !

PR #143, opened 2021-03-25
Fix warning after increasing to iOS 13

zkline101:


Sources/ArcGISToolkit/Extensions.swift
```diff
@@ -15,7 +15,7 @@ import Foundation
 import UIKit
 
 extension UIApplication {
-    func topViewController(_ controller: UIViewController? = UIApplication.shared.keyWindow?.rootViewController) -> UIViewController? {
+    func topViewController(_ controller: UIViewController? = UIApplication.shared.windows.filter { $0.isKeyWindow }.first?.rootViewController) -> UIViewController? {
```
philium, line 18: I think the whole point of `keyWindow` being deprecated is so that people _won't use_ the key window. However, if we must use it:

```suggestion
    func topViewController(_ controller: UIViewController? = UIApplication.shared.windows.first(where: { $0.isKeyWindow })?.rootViewController) -> UIViewController? {
```

It doesn't make much of a difference in this case because the number of windows is very small (probably one), but filtering first creates a new, intermediate array that is just thrown away immediately. It is always better to use `first(where:)` instead of `filter` and then `first`.

zkline101: ah yes I was just making that change

philium (approved):

mhdostal (approved):

PR #146, opened 2021-04-05
Add Toolkit Project 

zkline101:
Reasons for adding a usable toolkit project
- For internal reasons
- If users still wanna use the toolkit with the manually installed SDK

Let me know your thoughts if the project file  (and `info.plist`) should be in its own folder (maybe ArcGISToolkitProject/). Right now it is at the root of the repo. (Ting showed me an example that left the project file and .plist at the root)

Cc: @dg0yal @yo1995


zkline101:
This is no longer needed.

PR #155, opened 2021-08-23
Update package manifest for U12

zkline101:
Update 12 is not available for the SDK yet so heads up this will show an error. We don't need to merge this now but it should be ready to merge when we need to.


mhdostal (approved):

zkline101:
This looks like the last PR going into the `update12` branch so I am going to merge this.

PR #159, opened 2021-09-28
Chri7325/local

chri7325:
Issue-number: https://devtopia.esri.com/runtime/devops/issues/1250

This PR documents and adds small facilities to be able to use a local Toolkit references with a local Runtime for iOS installation to aid development and offline use cases.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -24,10 +24,12 @@
 
 import PackageDescription
 
-// To use a local installation of ArcGIS Runtime for iOS, uncomment the following lines and also uncomment
-// package depenencies:package:path and comment the depenencies:package:url
+// This is setup by default to use a hosted version of the ArcGIS Runtime for iOS.
+// To use a local installation of ArcGIS Runtime for iOS, uncomment the lines below declaring
+// the `localArcGISPackage` constant and use that as the package dependency instead of the
+// hosted version.
 // import class Foundation.ProcessInfo
-// let sdk_install_package: String = (ProcessInfo.processInfo.environment["HOME"] ?? "") + "/Library/SDKs/ArcGIS"
+// let localArcGISPackage = ProcessInfo.processInfo.environment["HOME"]! + "/Library/SDKs/ArcGIS"
 
 let package = Package(
     name: "arcgis-runtime-toolkit-ios",
@@ -43,7 +45,7 @@ let package = Package(
     dependencies: [
         .package(
             name: "arcgis-runtime-ios",
-            // path: sdk_install_package
+            // path: localArcGISPackage
             url: "https://github.com/Esri/arcgis-runtime-ios", .upToNextMinor(from: "100.12.0")
         )
     ],
--- a/Package.swift
+++ b/Package.swift
@@ -24,10 +24,10 @@
 
 import PackageDescription
 
-// To use an local installation of ArcGIS Runtime for iOS, uncomment the following lines and also uncomment
+// To use a local installation of ArcGIS Runtime for iOS, uncomment the following lines and also uncomment
 // package depenencies:package:path and comment the depenencies:package:url
 // import class Foundation.ProcessInfo
-// let sdk_install_package:String = (ProcessInfo.processInfo.environment["HOME"] ?? "") + "/Library/SDKs/ArcGIS"
+// let sdk_install_package: String = (ProcessInfo.processInfo.environment["HOME"] ?? "") + "/Library/SDKs/ArcGIS"
 
 let package = Package(
     name: "arcgis-runtime-toolkit-ios",
@@ -45,7 +45,7 @@ let package = Package(
             name: "arcgis-runtime-ios",
             // path: sdk_install_package
             url: "https://github.com/Esri/arcgis-runtime-ios", .upToNextMinor(from: "100.12.0")
-            )
+        )
     ],
     targets: [
         .target(
--- a/Package.swift
+++ b/Package.swift
@@ -24,6 +24,11 @@
 
 import PackageDescription
 
+// To use an local installation of ArcGIS Runtime for iOS, uncomment the following lines and also uncomment
+// package depenencies:package:path and comment the depenencies:package:url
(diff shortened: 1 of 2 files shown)


Package.swift
```diff
@@ -24,6 +24,11 @@
 
 import PackageDescription
 
+// To use an local installation of ArcGIS Runtime for iOS, uncomment the following lines and also uncomment
+// package depenencies:package:path and comment the depenencies:package:url
+// import class Foundation.ProcessInfo
+// let sdk_install_package:String = (ProcessInfo.processInfo.environment["HOME"] ?? "") + "/Library/SDKs/ArcGIS"
```
yo1995, line 30:
```suggestion
// let sdk_install_package: String = (ProcessInfo.processInfo.environment["HOME"] ?? "") + "/Library/SDKs/ArcGIS"
```
FYI last time Zach recommended using `import FileProvider` and `FileManager.default.homeDirectoryForCurrentUser.path`. I'll leave it to Zach's comments.

Package.swift
```diff
@@ -36,7 +41,11 @@ let package = Package(
         )
     ],
     dependencies: [
-        .package(url: "https://github.com/Esri/arcgis-runtime-ios", .upToNextMinor(from: "100.12.0"))
+        .package(
+            name: "arcgis-runtime-ios",
+            // path: sdk_install_package
+            url: "https://github.com/Esri/arcgis-runtime-ios", .upToNextMinor(from: "100.12.0")
+            )
```
yo1995, line 48:
```suggestion
        .package(
            name: "arcgis-runtime-ios",
            // path: sdk_install_package
            url: "https://github.com/Esri/arcgis-runtime-ios", .upToNextMinor(from: "100.12.0")
        )
```
Indentation is off here. Use <kbd>control</kbd> <kbd>i</kbd> in Xcode to auto-reindent

Package.swift
```diff
@@ -24,6 +24,11 @@
 
 import PackageDescription
 
+// To use an local installation of ArcGIS Runtime for iOS, uncomment the following lines and also uncomment
```
yo1995, line 27:
```suggestion
// To use a local installation of ArcGIS Runtime for iOS, uncomment the following lines and also uncomment
```

Package.swift
```diff
@@ -24,6 +24,11 @@
 
 import PackageDescription
 
+// To use a local installation of ArcGIS Runtime for iOS, uncomment the following lines and also uncomment
+// package depenencies:package:path and comment the depenencies:package:url
+// import class Foundation.ProcessInfo
+// let sdk_install_package: String = (ProcessInfo.processInfo.environment["HOME"] ?? "") + "/Library/SDKs/ArcGIS"
```
rolson, line 30:
```suggestion
// let localArcGISPackage = ProcessInfo.processInfo.environment["HOME"]! + "/Library/SDKs/ArcGIS"
```

snake case isn't common in swift. And then I think that `ProcessInfo.processInfo.environment["HOME"]!` should be safe to force unwrap. Also, we don't include the type when the compiler can infer it.

Package.swift
```diff
@@ -36,7 +41,11 @@ let package = Package(
         )
     ],
     dependencies: [
-        .package(url: "https://github.com/Esri/arcgis-runtime-ios", .upToNextMinor(from: "100.12.0"))
+        .package(
+            name: "arcgis-runtime-ios",
+            // path: sdk_install_package
```
rolson, line 46:
```suggestion
            // path: localArcGISPackage
```

Package.swift
```diff
@@ -24,6 +24,11 @@
 
 import PackageDescription
 
+// To use a local installation of ArcGIS Runtime for iOS, uncomment the following lines and also uncomment
+// package depenencies:package:path and comment the depenencies:package:url
```
rolson, line 28: These are not the technical way to refer to the swift functions. You could probably not include this information and make the instructions much more general and it would be acceptable.  Something like:

```suggestion
// This is setup by default to use a hosted version of the ArcGIS Runtime for iOS.
// To use a local installation of ArcGIS Runtime for iOS, uncomment the lines below declaring
// the `localArcGISPackage` constant and use that as the package dependency instead of the
// hosted version.
```

Package.swift
```diff
@@ -24,6 +24,11 @@
...
+// import class Foundation.ProcessInfo
```
chri7325, line 31: I'm wondering if we should even keep these commented out for an even easier ease of use? The variables will be unused but it may not really add that much overhead. Thoughts?

rolson: benefits of leaving them commented:
- the variable isn't used, so it's uncommented existence can be confusing
- if for some reason the force unwrap fails on some machine then leaving it commented won't cause any problems

mhdostal (approved):

rolson (approved):
I approve the code changes.  But @dg0yal needs to approve the product decision to have this.

dg0yal:
👍 I think this will be useful for the situations mentioned, while still not burdening typical usage

dg0yal (approved):

chri7325:
Thank you everyone. I think there's still one more pending question on the filesystem call but that is a bit out of my scope as I don't really know Swift. I did retest with all the changes and things are still working well. Would someone be able to merge this to move forward and follow up on the filesystem part or suggest a change here?

rolson:
What is the filesystem question?

chri7325:
@rolson, this is in reference to @yo1995 PR suggestion

> FYI last time Zach recommended using import FileProvider and FileManager.default.homeDirectoryForCurrentUser.path. I'll leave it to Zach's comments.

chri7325:
@yo1995 @zkline101 is there still any issues with the current implementation for checking the HOME environment variable or can this PR be merged now?

yo1995:
I don't have a preference and think the current solution is fine.

zkline101:
> I don't have a preference and think the current solution is fine.

Agreed. Also, tested locally worked fine for me, thanks!

mhdostal:
Thanks @chri7325!

PR #168, opened 2022-04-15
Fully set up the "siteFacilityPromptVC" before display.

mhdostal:
Fix a crash in the `siteFacilityPromptVC`.  The issue is that the view model was being set on the VC __after__ the VC was displayed.  This resulted in an incorrect initialization of the `filteredSearchFacilities` variable, which caused a crash in the `tableView(tableView:, didSelectRowAt:)` method.

Steps to reproduce:
- open FF example
- tap on "Sites" button
- tap on "Redlands Main"
- tap on "Building L" to select
- [map is displayed on Building L]
- tap on "Sites" button
- tap on "Redlands Main"
- tap on "filter" search bar to make the text field the first responder
- tap on "Building L".

Because `filteredSearchFacilities` was setup incorrectly (it has no elements), the code in `didSelectRowAt:` crashes with an "Index out of range" error.

The fix is to fully set up the `siteFacilityPromptVC` before displaying it.


dfeinzimer (approved):
Looks good. I was able to reproduce the crash on v.next and wasn't able to repro on this branch.

Issue #177, opened 2022-08-02
Audit upcoming release


dfeinzimer:
- [x] Compass
- [x] Measure
- [x] Scalebar
- [x] Legend
- [x] Job Manager
- [x] Time Slider
- [x] Popup Controller
- [x] Template Picker
- [x] AR (1 issue found #178 with resolution in PR #180)
- [x] Bookmarks
- [x] Floor Filter
- [x] Doc updates #179 
- [x] Run tests
- [x] Merge #179 
- [x] Merge #181 to address #176

PR #179, opened 2022-08-02
Update 15 Changes

dfeinzimer:
See #177


Documentation/TimeSlider/README.md
```diff
@@ -56,33 +56,33 @@ User can control whether to show `Slider` and/or `Playback Buttons` using `isSli
...
+timeSlider.labelMode = .ticks
+timeSlider.addTarget(self, action: #selector(TimeSliderExample.timeSliderValueChanged(timeSlider:)), for: .valueChanged)
+view.addSubview(timeSlider)
+
+// Add constraints to position the slider
+let margin: CGFloat = 10.0
+timeSlider.translatesAutoresizingMaskIntoConstraints = false
+timeSlider.bottomAnchor.constraint(equalTo: mapView.attributionTopAnchor, constant: -margin).isActive = true
+
+if #available(iOS 11.0, *) {
```
yo1995, line 70: IMO with the minimum required Runtime SDK version as 100.15.0, the min iOS version should be iOS 13, so this API check can be removed?

mhdostal: Nice catch!

Documentation/JobManager/README.md
```diff
@@ -7,35 +7,34 @@ The Job Manager is a toolkit component that you can just plug into the applicati
...
+	return
+    }
 
-            // make sure we are still around...
-            guard let strongSelf = self else {
-                return
-            }
+    // make sure task does not get released
+    guard let strongTask = task else{
+	return
```
zkline101, line 21: The indentation for this return statement looks off. Same with the guard statement above.

zkline101 (commented):
Thanks for cleaning up some of that doc! I have one small comment.

Documentation/JobManager/README.md
dfeinzimer: Good catch. Looks like there were tabs instead of spaces, and the wrong amount

zkline101 (approved):

PR #187, opened 2024-04-29
100.15.1 Release

zkline101:
We can merge this when the `100.15.5` release of the iOS SDK is released. @dg0yal confirmed this release version for the toolkit. Also, I don't think we need to update [this](https://github.com/Esri/arcgis-runtime-toolkit-ios/blob/a729b0660068dc06446cd90fabdbf0723c8131b1/Package.swift#L49) line since it looks like that pulls down that latest minor and patch release of the iOS SDK so it should automatically get the correct version.


mhdostal (changes requested):
You should also set the `MARKETING_VERSION` to `100.15.5` [here](https://github.com/Esri/arcgis-runtime-toolkit-ios/blob/a729b0660068dc06446cd90fabdbf0723c8131b1/Examples/ArcGISToolkitExamples.xcodeproj/project.pbxproj#L413C5-L413C22) and [here](https://github.com/Esri/arcgis-runtime-toolkit-ios/blob/a729b0660068dc06446cd90fabdbf0723c8131b1/Examples/ArcGISToolkitExamples.xcodeproj/project.pbxproj#L434).

zkline101:
> You should also set the MARKETING_VERSION to `100.15.5`

Do you mean `100.15.1` ? Since this is referring to the new version of the toolkit not the SDK.

I'll open another PR for this.

zkline101:
> Usually we always have the Toolkit and SDK version match, 

Yeah for sure this is a slightly different direction. But Divesh has confirmed this version number.

mhdostal (approved):

PR #189, opened 2024-05-03
Move privacy manifest file location

zkline101:
Cocoapods didn't like the original location of the privacy manifest so it wouldn't be picked up the privacy report. This fixes that. Thanks @saratchandrakarumuri for testing this.

I also confirmed this change still allows the privacy manifest to be picked up when installing the toolkit via SPM.


saratchandrakarumuri (approved):
Verified the new location with cocoapods and manifest is showing up as expected

mhdostal (approved):

